PR #21, opened 2020-08-08
Solution

rmnkk:


src/makeCalculator.js
```diff
@@ -35,7 +35,53 @@
...
+      return this;
+    },
+
+    add(res, value) {
+      let operationResult = res;
+
+      operationResult += value;
+
+      return operationResult;
+    },
```
DmJSCourse, line 53: add(res, value) {
       return res + value;
    },

rmnkk: Thank you for your advice! 👍

PR #23, opened 2020-08-09
solution

oggyrock:


src/makeCalculator.js
```diff
@@ -35,7 +35,47 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, value) {
+      this.result = operation(value);
```
furude-rika, line 48: I'm not sure but maybe you can omit the assigning and use only 'operation(value)', it will work the same

oggyrock: Looks good, thanks!

PR #26, opened 2020-08-09
Develop

oivannikov:


src/makeCalculator.js
```diff
@@ -34,8 +34,48 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, num) {
+      calculator.result = callback(num);
```
irenhh, line 73: assignment is not necessary, just call the callback

PR #33, opened 2020-08-10
solution

masyatya:


src/makeCalculator.js
```diff
@@ -35,7 +35,47 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, value) {
+      calculator.result = callback(value);
```
irenhh, line 72: just calling a callback is enough

PR #34, opened 2020-08-10
solution

furude-rika:


src/makeCalculator.js
```diff
@@ -35,7 +35,47 @@
...
+    divide(num) {
+      calculator.result /= num;
+
+      return calculator.result;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
```
irenhh, line 74: be consistent, use either `this` or the name of the object

PR #40, opened 2020-08-10
solution

artsashko:


src/makeCalculator.js
```diff
@@ -35,7 +35,44 @@
...
+        case '-':
+          this.result -= value;
+          break;
+        case '*':
+          this.result *= value;
+          break;
+        case '/':
+          this.result /= value;
+          break;
+      }
```
irenhh, line 70: don't you think writing logic in corresponding methods would be simpler?
don't forget about default case

PR #44, opened 2020-08-10
Solution

nznmv:


src/makeCalculator.js
```diff
@@ -35,7 +35,46 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, num) {
+      calculator.result = callback(num);
```
irenhh, line 72: you don't really need this assignment

irenhh: just calling the callback is enough

PR #45, opened 2020-08-10
solution

Oleg-Ozerov:


src/makeCalculator.js
```diff
@@ -36,6 +36,41 @@
...
+
+      return calc.result;
+    },
+    reset() {
+      calc.result = 0;
+
+      return calc;
+    },
+    operate(callback, number) {
+      calc.result = callback(number);
```
irenhh, line 67: there is no need to assign the result of callback to calc.result again as you do this in the callback

PR #51, opened 2020-08-13
solution

VenomSRT:


src/makeCalculator.js
```diff
@@ -35,7 +35,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      calculator.result += number;
```
irenhh, line 54: be consistent, either use `this` or the name of the object

PR #52, opened 2020-08-14
calc

yuriimovchko:


src/makeCalculator.js
```diff
@@ -35,7 +35,43 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
+    result: 0,
+
+    operate(operator, num) {
+      calc.result = operator(num);
```
irenhh, line 42: no need to do an assignment, you already have it in the corresponding methods

PR #133, opened 2020-12-17
solution

chri911:


MarchD (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
```
vsm2k17, line 42:
```suggestion
    result: 0,
    
    operate(callback, number) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      let result = calculator.result;
```
vsm2k17, line 55:
```suggestion
      calculator.result += number;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+
+      result += number;
+
+      return result;
```
vsm2k17, line 59:
```suggestion
      return calculator.result;
```

vsm2k17: don't return some computed value, use the field "result" of your calculator object

vsm2k17 (changes requested):

src/makeCalculator.js
chri911: fixed

vsm2k17 (approved):

PR #190, opened 2021-02-25
Solution

gennadytabala:


src/makeCalculator.js
```diff
@@ -37,7 +37,69 @@
...
+  //   return result - number;
+  // };
+
+  // calculator.multiply = function(result, number) {
+  //   return result * number;
+  // };
+
+  // calculator.divide = function(result, number) {
+  //   return result / number;
+  // };
```
vkryvytskyy, line 100: remove comments

src/makeCalculator.js
```diff
@@ -37,7 +37,69 @@
...
+
+    subtract: function(result, number) {
+      return result - number;
+    },
+
+    multiply: function(result, number) {
+      return result * number;
+    },
+
+    divide: function(result, number) {
```
vkryvytskyy, line 68: you don't need to pass result into this function, only number

vkryvytskyy (changes requested):

gennadytabala:
It was one of previous solutions :)
When we have "add(result, number) ...", we don't need "this" inside add, substract and others. So, this functions is "clear".
It is more comlpicated but external interface the same - complexity inside.

PR #225, opened 2021-04-26
solution

Ihor-Karpyn:


src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    multiply(n) {
+      calc.result *= n;
+    },
+    divide(n) {
+      calc.result /= n;
+    },
+    reset() {
+      calc.result = 0;
+
+      return calc;
```
IvanGrekov, line 58: It's not a mistake, but I think it's not adaptive. For my opinion, more flexible will be if we return `this` instead of name of our object

IvanGrekov (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
```
mykola-yarchuk, line 40:
```suggestion
  return {
```

src/makeCalculator.js
mykola-yarchuk, line 58:
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    },
+    reset() {
+      calc.result = 0;
+
+      return calc;
+    },
+    operate(callback, number) {
+      callback(number);
+
+      return calc;
```
mykola-yarchuk, line 63:
```suggestion
      return calc;
```
```suggestion
      return this;
```

mykola-yarchuk (approved):

PR #230, opened 2021-04-27
Solution v0.1

SkyPKWolf:


igor-stasiv (approved):
Good job! But don't name variables from the upper case letter, because it means that the variable is a constructor or class. And I'm not sure 100% about that, so I approve it and leave it for mentor review.

mykola-yarchuk (approved):
great

PR #233, opened 2021-04-27
solution

provolovskyi-illia:


Ihor-Karpyn (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
```suggestion
  return {
```
You can directly return the object

varseniuk (approved):
Good job! 👍

PR #235, opened 2021-04-28
solution

vitalya2626:


Ihor-Karpyn (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
```suggestion
  return {
```
You may directly return the object

varseniuk (approved):
Good job! 👍

PR #240, opened 2021-04-28
solution

RuslanaPodshyvalova:


src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const calculatorObj = {
```
Ihor-Karpyn, line 41:
```suggestion
  const calculator = {
```

Ihor-Karpyn (approved):

src/makeCalculator.js
varseniuk: Ihor is right. Let's not use `obj` in variables names

varseniuk (approved):
Great job! 🔥

PR #243, opened 2021-04-28
solution

dmitry-puhliakov:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
```
vova-myronenko, line 40: it's up to you, but you could return the object right away without declaring a variable
```suggestion
  return {
```

vova-myronenko (approved):

src/makeCalculator.js
vova-myronenko: Anyway, well done!

varseniuk: Yep. Good advice from Vova

varseniuk (approved):
Nice 🔥

PR #256, opened 2021-05-04
Solution

Kirill-Osypov:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      calculator.result /= value;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(f, value) {
```
vkryvytskyy, line 65:
```suggestion
    operate(method, value) {
```
give meaningful name to functions

vkryvytskyy (approved):
good job

PR #258, opened 2021-05-12
Solution-001

senkiv-oleh:


src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
-  // write code here
+function makecalculator() {
+  const calc = {
+
+    result: 0,
+
+    operate(callback, value = 0) {
+      callback(value);
+
+      return calc;
```
varseniuk, line 47:
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+    },
+
+    add(value = 0) {
+      calc.result += value;
+
+      return calc;
```
varseniuk, line 53:
```suggestion
```
It's redundant here. As task description says:

> - The `operate` and `reset` methods can be called in a chain.

So it means that only `reset` and `operate` should return the calculator to have the possibility to be used in the chain.
But ok

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+    divide(value = 0) {
+      calc.result /= value;
+
+      return calc;
+    },
+
+    reset() {
+      calc.result = 0;
+
+      return calc;
```
varseniuk, line 77:
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
  *
  * @return {object}
  */
-function makeCalculator() {
-  // write code here
+function makecalculator() {
+  const calc = {
```
```suggestion
  return {
```
You can return this object right away

varseniuk (approved):
Good!
Left some comments

PR #265, opened 2021-05-30
Solution

Yaslik1033:


src/makeCalculator.js
```diff
@@ -36,8 +36,35 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const calc = {
+    result: 0,
+    add(x) {
+      this.result += x;
```
vkryvytskyy, line 44:
```suggestion
      calc.result += x;
```
you can use calc to avoid `.call` and this manipulations

vkryvytskyy (approved):

PR #328, opened 2021-08-07
solved

lera-yaskevych:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
alex-glotov67, line 40: You can just write return without object declaration:
```suggestion
return {
    result: 0,
```

alex-glotov67 (approved):
Nice solution 🔥 Left some advice

src/makeCalculator.js
lera-yaskevych: Good point, thanks)

VinnieJ-2k20 (approved):
Good job 👍

PR #338, opened 2021-08-09
solution

DmytroMykoliv:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
```
Vova-Stoliar, line 61: You can add throw new Error('some text').  if (b === 0)

Vova-Stoliar (approved):
Good job 🔥

src/makeCalculator.js
DmytroMykoliv: Ok, thanx, good advise  ))

herkoss (approved):
Well done!👍👏

PR #344, opened 2021-08-10
solution calculator done

AndreySerhovetc:


evgen1sider (approved):
this cool

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(number, result) {
```
VinnieJ-2k20, line 43: Need to use more appropriate variable names.
What you accept as an argument - is not a result.
It can be `number1` and `number2` or `firstNumber` and `secondNumber`

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      return result + number;
+    },
+
+    subtract(number, result) {
+      return result - number;
+    },
+
+    divide(number, result) {
+      if (result === 0) {
+        throw new Error('value is not correct');
```
VinnieJ-2k20, line 53: Try to explain errors more
Also, you divide by `number`, but check `result` for being 0.

```suggestion
        throw new Error('value is not correct - division by 0');
```

VinnieJ-2k20 (changes requested):
Almost done.
Just need to fix some variable names and error check logic.

src/makeCalculator.js
AndreySerhovetc: fixed

VinnieJ-2k20 (approved):
Good job 👍

PR #347, opened 2021-08-10
first commit

VVHimiak:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(number1, number2) {
```
sasha-morozov, line 42: It looks good, but I would suggest just using "a, b" as variables. By the name of methods, it's already clear, that you will pass integers as parameters.

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+      return (number1 + number2);
+    },
+    subtract(number1, number2) {
+      return (number1 - number2);
+    },
+    multiply(number1, number2) {
+      return (number1 * number2);
+    },
+    divide(number1, number2) {
```
sasha-morozov, line 51: If the "number2" equals 0, the function should throw an error.

sasha-morozov (approved):
Overall, the code looks good. However, I recommend adhering to the suggestions I left above)

src/makeCalculator.js
VVHimiak: Thanks! I appreciate your review and I'll fix the mistakes that you mentioned!

herkoss (approved):
Good job!👏👍

PR #348, opened 2021-08-10
calculator

zikozakazabiyaka:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
sasha-morozov, line 40: You may use return instead of declaring a variable.

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+    add(a, b) {
+      return a + b;
+    },
+
+    subtract(a, b) {
+      return a - b;
+    },
+
+    divide(a, b) {
```
sasha-morozov, line 52: We can't divide by zero. So make sure the method will work correctly if the second parameter is '0'.
Just read about 'throw' on MDN)

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    },
+
+    operate(callback, value) {
+      this.result = callback(this.result, value);
+
+      return this;
+    },
+  };
+
+  return calculator;
```
sasha-morozov: If you adhere to the suggestion above, you will no longer need this return

sasha-morozov (approved):
The code is cool. Left some suggestions above

src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
+      return a + b;
+    },
+
+    subtract(a, b) {
+      return a - b;
+    },
+
+    divide(a, b) {
+      if (b === 0) {
+        throw new Error('Error');
```
VinnieJ-2k20, line 54: Try to describe an error in more detailed way
```suggestion
        throw new Error('Division by 0');
```

VinnieJ-2k20 (approved):
Good job 👍

PR #358, opened 2021-08-11
done

Uran-web:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
 
+      return this;
+    },
+
+    add(value) {
+      return this.result + value;
+    },
+
+    divide(value) {
+      if (value !== 0) {
```
Clonvhs, line 61: delete it we
we can divide at 0 and result will be 0 always

Clonvhs (changes requested):

src/makeCalculator.js
Uran-web: Thank you for your review! I absolutely agree - code will be work, but I think, it would be better for continuously using.

PR #375, opened 2021-09-17
solution

vmiskiv:


CouldSweeT (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return a - b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
```
aliceblunt8, line 57: pass just value from the operate arguments (num), no need to pass the result, you can get it inside the operation itself

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    operate(callback, num) {
+      this.result = callback(this.result, num);
```
aliceblunt8, line 60:
```suggestion
      callback(num);
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #388, opened 2021-09-21
solution 🧮

caelum29:


src/makeCalculator.js
```diff
@@ -37,7 +37,32 @@
...
+    divide(a, b) {
+      return a / b;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(operation, value) {
+      this.result = operation(this.result, value);
```
aliceblunt8, line 61: pass just value 

```suggestion
    operate(operation, value) {
      operation(value);
```

aliceblunt8 (changes requested):

src/makeCalculator.js
caelum29: Are you sure?? 🤔

aliceblunt8 (approved):

PR #391, opened 2021-09-21
Solution

Denis-Demyanok:


Anna-Tkachenko (approved):
Good job!👍
You can also try to rewrite logic using 'this' keyword

PR #395, opened 2021-09-22
solution

KatyKuzmenko:


Anna-Tkachenko (approved):
Good job!👍
You can also try to rewrite logic of math methods using 'this' keyword

PR #428, opened 2021-11-10
Make a calculator

dmtrhrytsak:
You're not a programmer if you haven't made a calculator at least once (⌐■_■)

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(x) {
+      this.result += x;
+    },
+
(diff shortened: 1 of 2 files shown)


illiarerun (approved):
LGTM 🔥

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    subtract(x) {
+      this.result -= x;
+    },
+
+    multiply(x) {
+      this.result *= x;
+    },
+
+    divide(x) {
+      this.result /= x;
```
illiarerun, line 56: But what would be if the result is 0?))

dmtrhrytsak: self-destruction :D so-called kamikaze

PR #442, opened 2021-11-12
Calculator

Vitaliy-0:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj =  {
```
alex-kostomarov, line 40: i'd call it calculator)

alex-kostomarov (approved):
Great job!

ihor-jpeg (approved):
Very good work!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      calculator.result -= value;
+    },
+
+    multiply(value) {
+      calculator.result *= value;
+    },
+
+    divide(value) {
+      calculator.result /= value;
+    },
```
ihor-jpeg, line 69: dividing by zero in JS will not throw an error, it will yield Infinity or -Infinity. be aware of that

PR #443, opened 2021-11-13
Solution

alex-kostomarov:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, num) {
+      this.result = callback.call(this, num);
```
Vitaliy-0, line 44: you can just call callback like
callback(value)

Vitaliy-0 (approved):
GJ

aionova (approved):
GJ!

PR #445, opened 2021-11-13
solution

MagicBox15:


kurtkobein (approved):
You don't have to return the result of each operation. You can use bind, call or apply methods to set object context in each operation and change result field in operation.

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    Methods: [`add`, `subtract`, `multiply`, `divide`, `reset`, `operate`],
```
aionova, line 41: why do you need this property?

aionova (approved):

src/makeCalculator.js
MagicBox15: when I was solving the task, I thought that I have return possible operations 😕

PR #447, opened 2021-11-14
Solution

olya-shyian:


alinaShchavinska (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return a * b;
+    },
+
+    subtract(a, b) {
+      return a - b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```
ihor-jpeg, line 57: dividing by zero will yield infinity. JS doesn't throw an error in this case, but you should not allow it.

ihor-jpeg (approved):

ihor-jpeg (approved):
Good work!

PR #450, opened 2021-11-14
Solution

volodymyr-soltys97:


skachkov23 (approved):
LGTM!

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    operate(action, number) {
+      this.result = action(this.result, number);
```
ihor-jpeg, line 44: i think you are doing an unnecessary action here. all the methods have direct access to your result variable. why pass it as an argument?

ihor-jpeg (approved):
Good work! Have a look at my comment

PR #453, opened 2021-11-14
Solution

4qwerty:


DudarDaryna (approved):
GJ

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```
ihor-jpeg, line 63: This will not throw an error in case of b = 0. It will return Infinity or -Infinity. Keep that in mind

ihor-jpeg (approved):
Good work!

PR #456, opened 2021-11-15
add calculator

i-stanko:


tmrsl (approved):
Looks good

src/makeCalculator.js
```diff
@@ -36,8 +36,36 @@
...
+      return calculator;
+    },
+    operate(callback, number) {
+      callback(number);
+
+      return calculator;
+    },
+
+    add(number) {
+      calculator.result += number;
```
ihor-jpeg, line 55: you can refer to the `calculator` as `this` to make it clear that the `result` is taken from the context

ihor-jpeg (approved):
Well done!

PR #459, opened 2021-11-15
solution: advanced_calculator task

lanebx:


tmrsl (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+
+    operate(callback, value) {
+      callback(value);
+
+      return this;
+    },
+
+    add(value) {
+      calculator.result += value;
+    },
```
ihor-jpeg, line 57: it's best to use `this` instead of `calculator`, so that it is clear that the `result` is taken from the context

ihor-jpeg (approved):
Good work!

PR #461, opened 2021-11-16
Solution

aldmpr:


Veronika-Korona (approved):
GJ!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    divide(number) {
+      return calculator.result / number;
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
```
ihor-jpeg, line 64: What is the difference here between calculator.result and this.result?

ihor-jpeg (changes requested):
If I decide to name my calculator differently i suddenly need to change that name in many places.

aionova (approved):
GJ!

PR #487, opened 2021-12-20
solution

mykola-koval:
but why calculations are not working with 'this' ?

`calculator.result += number;`
gives me
` TypeError: Cannot read property 'result' of undefined`

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(number) {
+      calculator.result += number;
+
+      return this;
+    },
+    subtract(number) {
+      calculator.result -= number;
+
+      return this;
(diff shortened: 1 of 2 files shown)


bcmp4p1 (approved):

YevheniiKorzun (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(number) {
+      calculator.result += number;
```
danheim, line 43: use this

danheim (changes requested):

src/makeCalculator.js
mykola-koval: i cannot use 'this' here. because of using 'operate' method this returns undefined.

danheim (approved):

PR #489, opened 2021-12-20
Solu tion

destroythemoon:


src/makeCalculator.js
```diff
@@ -37,7 +37,29 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => a / b,
+
+    operate: function(action, number) {
```
HrynykNataliia, line 48: checkist - if you creating a method in the object, you don't need to use function keyword, use shortcut instead.

src/makeCalculator.js
```diff
@@ -37,7 +37,29 @@
...
+      this.result = action(this.result, number);
+
+      return this;
+    },
+
+    reset: function() {
```
HrynykNataliia, line 54: тут те саме

HrynykNataliia (changes requested):

PR #497, opened 2021-12-20
Solution

AlexAzalor:


levix2l (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+    divide(elem) {
+      calculator.result /= elem;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+    result: 0,
```
Elena-brv, line 62: add result to the start of the object, it's difficult to understand now, what the result on all these methods

Elena-brv (approved):

PR #543, opened 2022-02-08
solution

Tislenok:


src/makeCalculator.js
```diff
@@ -36,8 +36,42 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```
aliceblunt8, line 71: let's change the result inside each of these methods, then operate will just call the callback with number

aliceblunt8 (dismissed):

PR #549, opened 2022-02-14
Solution

Ol-Lav:


krolich500grn (approved):
i'm not sure that we need return calculator in this task, but it's not a mistake
and in my opinion we need return this in reset method
but it's only my subjective opinion :)

solaryasha (approved):

PR #554, opened 2022-02-15
solution

ghost:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return this;
+    },
+
+    operate(callback, n) {
+      callback.call(this, n);
+
+      return this;
+    },
+
+    add(n) {
```
krolich500grn, line 55: n is bad name for property, use number here

krolich500grn (changes requested):
good job, all you need is another property name

krolich500grn (approved):

solaryasha (approved):

PR #607, opened 2022-04-07
solution2

barannikovArtem:


YegorKochetkov (approved):

YegorKochetkov (changes requested):
it seems to me that for proper operation it is necessary to use `this.result` instead of `calc.result`

if we create one more calculator calculations in one will affect the result in the other

Andr1yK (approved):

PR #609, opened 2022-04-08
calculator

Nikitiliok:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
```
YegorKochetkov, line 40: ojb is uninformative name. use calc or calculator instead of

YegorKochetkov (changes requested):
it seems to me that for proper operation it is necessary to use this.result instead of obj.result

if we create one more calculator calculations in one will affect the result in the other

MaksimusCRS (approved):

PR #624, opened 2022-04-27
solution

efdalyalcin:


LukaszGryziewicz (approved):
Good job 🥇. I think the 'this' keyword doesn't seem to work in these functions because of the way the tests are written. The object that we create also can't be returned outright and has to be named 'calculator' because of the same reason.

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #628, opened 2022-04-30
solutionI

EnjSadman:


src/makeCalculator.js
```diff
@@ -38,6 +38,49 @@
...
+    result: 0,
+
+    add: (result, b) => result + b,
+    multiply: (result, b) => result * b,
+    divide: (result, b) => result / b,
+    subtract: (result, b) => result - b,
+
+    operate: function(operation, b = 0) {
+      switch (operation) {
+        case this.add: {
```
alexandra-protyanova, line 53: I guess you don't really need all that cases, they're a bit unnecessary)

alexandra-protyanova (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,29 @@
...
   // write code here
+
+  const calculator = {
+
+    result: 0,
+
+    add: (result, b) => result + b,
+    multiply: (result, b) => result * b,
+    divide: (result, b) => result / b,
+    subtract: (result, b) => result - b,
```
Anna-Tkachenko, line 49: This functions maybe don't know about some result
Will be better to name arguments as a, b (just some two values for calculations)

Anna-Tkachenko (approved):
⭐️

PR #631, opened 2022-05-01
solution

iLebo-stack:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
```
rHashTag, line 40: Like we discussed in class, we need to be obvious when naming our variables. `obj` seems too generic in this case.

rHashTag (changes requested):
Good job with the task!

I've left a comment regarding the coding style.

rHashTag (approved):
LGTM!

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #633, opened 2022-05-01
solution

ZUBOV-ILLIA:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
```
valerii-derkach, line 40: better use full word 'object' to name variables

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+      obj.result /= x;
+    },
+
+    reset() {
+      obj.result = 0;
+
+      return obj;
+    },
+
+    operate(x, y) {
```
valerii-derkach, line 65: How the calculator will work:
 * - Each `operate` call takes a callback and a number and **sets the
 *   appropriate value to the `result` property.**
in your case the result property is changing in methods: `add`, `subtract`, `multiply`, `divide`, but not in 'operate'

valerii-derkach (changes requested):

src/makeCalculator.js
Anna-Tkachenko: What is obj?

Anna-Tkachenko (approved):

PR #634, opened 2022-05-01
solution

yelnikov-andrii:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(func, value) {
```
valerii-derkach, line 42: How the calculator will work:
 * - Each `operate` call takes a callback and a number and **sets the
 *   appropriate value to the `result` property.**
in your case the result property is changing in methods: `add`, `subtract`, `multiply`, `divide`, but not in 'operate' method

valerii-derkach (changes requested):

Anna-Tkachenko (changes requested):
Fix @valerii-derkach comment

yelnikov-andrii:
> Fix @valerii-derkach comment

I don't understand this problem, all tests are working

Anna-Tkachenko (approved):
Good job!

PR #635, opened 2022-05-01
Calculator was done

vlasiuk-anatolii:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
```
valerii-derkach, line 44: callback should  returns only the result of operations, but not re-asighns the calculator result

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    multiply(number) {
+      calculator.result *= number;
+    },
+
+    divide(number) {
+      calculator.result /= number;
+    },
+
+    operate(callback, number) {
+      callback(number);
```
valerii-derkach, line 60: this method should asigns the calculator result from callback

valerii-derkach (changes requested):

Anna-Tkachenko (changes requested):
Fix @valerii-derkach comments

Anna-Tkachenko (approved):
⭐️

PR #641, opened 2022-05-02
solution

SavMary:


valerii-derkach (approved):
good job, just one remark about naming of variables - better use full word 'object' instead 'obj'

src/makeCalculator.js
```diff
@@ -16,28 +16,53 @@
...
+    },
+    multiply: (result, number) => {
+      return result * number;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(f, number) {
```
Anna-Tkachenko, line 58: What is f?

Anna-Tkachenko (approved):
Good job! 👍

PR #642, opened 2022-05-02
adv calc

elkhayate:


59MrRobot (approved):
👏 👏

src/makeCalculator.js
```diff
@@ -38,6 +38,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      callback(number, this);
```
ihor-jpeg, line 67: you can do that in a cleaner way. the reason you need `this` is because you need to pass the context. you can do something like `callback.call(this, number)` and you no longer need `calculator` as a parameter of each operation function, simply use this. 

it's just a suggestion

ihor-jpeg (approved):

src/makeCalculator.js
elkhayate: thank you

VinnieJ-2k20 (approved):

PR #647, opened 2022-05-03
solution

WojciechCzarnowski:


andreisucman (approved):
Nice and clean

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
VinnieJ-2k20, line 40: Could simply `return` it right away and use `this` in methods,

VinnieJ-2k20 (approved):
Good job 👍 
Added one comment.

PR #666, opened 2022-05-14
Solution

07Spirit:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
-  // write code here
+  Methods.reset();
+
+  return Methods;
 }
 
+const Methods = {
+  result: 0,
+  operate: function(a, b) {
+    a(b);
```
aliceblunt8, line 48:
```suggestion
  operate: function(callback, number) {
    callback(number);
```

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  Methods.reset();
+
+  return Methods;
 }
 
+const Methods = {
```
aliceblunt8, line 45: why upper case? just methods?
```suggestion
const calculator = {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  Methods.reset();
```
aliceblunt8, line 40: just create an object inside makeCalculator and then return it
this way each time you will call makeCalculator it will create a new object in memory

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
 
+const Methods = {
+  result: 0,
+  operate: function(a, b) {
+    a(b);
+
+    return Methods;
+  },
+  add: function(b) {
+    Methods.result += b;
```
aliceblunt8, line 53:
```suggestion
    calculator.result += b;
```

07Spirit:
I did it. :smiley:

Bohdan-Maksymiuk (approved):

PR #681, opened 2022-06-03
solution

frontend112:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    operate(operation, number) {
+      this.result = operation(this.result, number);
```
meforoma, line 43: you can just execute the callback (operation in your case) on a number
after you adjust methods a bit (see further comments)
```suggestion
      operation(number);
```

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+
+      return this;
+    },
+    add(total, number) {
+      return number + total;
+    },
```
meforoma, line 49: for every operation you can interact directly with `this.result` and a `number` - so you'll need only one parameter
```suggestion
    add(number) {
      this.result += number;
    },
```

meforoma (changes requested):

src/makeCalculator.js
frontend112: Firstly I tried with one parametr but got `this` as undefined, 
forgot `this` depends on context runtime 
so each operation method don't have access to `this`

meforoma (approved):

PR #687, opened 2022-06-06
Advanced calculator

vsupruniuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(oparation, number) {
```
illiarerun, line 43:
```suggestion
    operate(operation, number) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      calculator.result += number;
+    },
```
illiarerun, line 57: right now we can execute this code:
image
i guess it should work a bit other way, we should change object that we called `operate` on, operation shouldn't be dependant on exact object, you can somehow pass needed object from operate to this actual `operation`

illiarerun (changes requested):
GJ, now let's work on fixes:

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+    result: 0,
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
+    },
+
+    reset() {
+      calculator.result = 0;
```
illiarerun, line 50: use `this` here also, issue is still there

illiarerun (changes requested):

jstmpelowycz (approved):

PR #696, opened 2022-06-09
Solution

stanislav-topikha:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value = 0, obj = this) {
+      obj.result += value;
```
jstmpelowycz, line 44: here, you shouldn't refer to `obj`, try `this`

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value = 0, obj = this) {
```
jstmpelowycz, line 43: you don't need to set a default value as `this` for `obj`

jstmpelowycz (changes requested):
know what a problem you faced, try [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) one

stanislav-topikha:
> know what a problem you faced, try [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) one and [this](https://www.freecodecamp.org/news/what-to-do-when-this-loses-context-f09664af076f/) one

Yep, that helped me a lot)

src/makeCalculator.js
stanislav-topikha: fixed

src/makeCalculator.js
stanislav-topikha: fixed

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, value) {
+      return operation.bind(this)(value);
```
Olexandr6, line 74: 👍

Olexandr6 (approved):
Well done🔥

PR #697, opened 2022-06-09
Solution

liliia-livandovskaya:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+
+    result: 0,
+    add(a, b) {
```
jstmpelowycz, line 43:
```suggestion
  const calculator = {
    result: 0,

    add(a, b) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+      return a + b;
+    },
+    subtract(a, b) {
+      return a - b;
+    },
```
jstmpelowycz, line 48:
```suggestion
    subtract: (a, b) => a - b;
```
to make it look simpler :)

jstmpelowycz (approved):
LGTM

src/makeCalculator.js
liliia-livandovskaya: Thank you! :)

PR #726, opened 2022-06-14
Solution

olya-yevsieienko:


AKomashchuk (approved):
Молодець )    
 Тільки не розумію навіщо ти добавляла  в divide :
      if (b === 0) {
        return `Incorrect value #{b}`;
      }

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      if (b === 0) {
+        return `Incorrect value #{b}`;
+      }
```
Olexandr6, line 64: 🔥

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      if (b === 0) {
+        return `Incorrect value #{b}`;
```
Olexandr6, line 63: Let's make this message more informative. Cause for someone, it's not clear why 0 is an incorrect value actually, `0` is correct, but we can't divide by 0

Olexandr6: And you should throw error in such case

Olexandr6 (changes requested):
Good job👍 Let's fix the error message

olya-yevsieienko:
Thanks, I changed that!

вт, 14 июн. 2022 г. в 14:45, Olexandr Yatsykovych ***@***.***

Olexandr6 (approved):
🔥

PR #738, opened 2022-06-21
solution

LAndrikevych:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    subtract(result, value) {
+      return result - value;
+    },
+
+    multiply(result, value) {
+      return result * value;
+    },
+
+    divide(result, value) {
+      return result / value;
```
illiarerun, line 62: what if value is `0`? can we divide by `0`?

illiarerun (approved):
GJ, left one comment

PR #751, opened 2022-07-04
Solution

Blankevich:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+    },
 
+    multiply(a, b) {
+      return a * b;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
```
valerii-derkach, line 67: you should use `this.result` instead of `calculator.result` and return `this `instead of `calculator`, when you code methods

valerii-derkach (approved):
Good job! 
just have some small fix to do in comment

PR #755, opened 2022-07-04
Solution

alexbardyshev:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    divide: (x, y) => {
+      return x / y;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+    operate(operation, number) {
+      calculator.result = operation(calculator.result, number);
```
solaryasha, line 60: Smart. Also, you could try to use `call`,`apply` or `bind` method, if you would like to call operation function without `calculator` reference

solaryasha (approved):

PR #762, opened 2022-07-07
initial solution

DashaPoliazz:


VETALEK (changes requested):
Good start 🔥

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      calculator.result += number;
```
VETALEK, line 57:
```suggestion
      this.result += number;
```

you need to use `this` instead of `calculator`

VETALEK (changes requested):
Almost 💪

src/makeCalculator.js
DashaPoliazz: Yeah, I tryed using `this `in callback methods `add, subtract `etc. But сontext `this `inside the callbacks equals undefined. I tryed `bind `context, but smth went wrong ((

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      this.result += number;
```
ivan-kadykalo, line 57: Use pure function, return the result and then save it only throw operate method
```
add(num1, num2) {
  return num1 + num2;
}

PR #765, opened 2022-07-07
Added simple calculator

LordAsmodey:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, num) {
+      callback.call(this, num);
+
+      return this;
+    },
```
VETALEK, line 69: it won't work if I use some artificial callback, like
`calculator.operate((a, b) => (a + b), 42)`

let's do like this
```suggestion
    operate(callback, num) {
      this.result = callback.call(this, this.result, num);

      return this;
    },
```

VETALEK (changes requested):
GJ 🚀 
commented

Ihor-Karpyn (approved):
LGTM

PR #773, opened 2022-07-11
solution

denys-danyliuk:


nikita-biriukov (approved):
🚀 🚀

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+  const calculator = {
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+
+    add(number) {
+      calculator.result += number;
+    },
```
ivan-kadykalo, line 49: Make it like a pure function.
Let this method return you a pure result, and save it in your `operate` method.
For example
```
add(num1, num2) {
	return num1 + num2;
}

File: src/makeCalculator.js [hunk 2]
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    operate(callback, number) {
+      callback(number);
```
ivan-kadykalo, line 42: Save your result here
`this.result = callback ...`

ivan-kadykalo (changes requested):

src/makeCalculator.js
denys-danyliuk: got it👍

ivan-kadykalo (approved):
🚀

PR #776, opened 2022-07-11
solution

AleksandrYuschenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
+      this.result = callback.call(this, this.result, number);
+
```
chernobrovin, line 44:
```suggestion
      this.result = callback.call(this.result, number);

```

I think this value is redundant here

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+    subtract(result, number) {
+      this.result -= number;
+
+      return this.result;
+    },
+    multiply(result, number) {
+      this.result *= number;
+
+      return this.result;
```
chernobrovin, line 65: why do you keep doing the same thing: `return this.result;` over and over? 

why don't you do it 1 time at the end of all calculations?

chernobrovin (changes requested):

chernobrovin (commented):
Everything is good, but you need a little bit "clean" your code

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(result, number) {
+      this.result += number;
+
+      return this.result;
+    },
```
Ihor-Karpyn, line 56: fix in all cases
```suggestion
    add(a, b) {
      return a + b
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
+      this.result = callback.call(this, this.result, number);
+
+      return this;
+    },
```
Ihor-Karpyn, line 46:
```suggestion
    operate(callback, number) {
      this.result = callback(this.result, number);

      return this;
    },
```

src/makeCalculator.js
Ihor-Karpyn: agree

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      return this.result;
+    },
+    divide(result, number) {
+      this.result /= number;
+
+      return this.result;
+    },
+  };
+
+  return calculator;
```
Ihor-Karpyn, line 74:
```suggestion
```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
Ihor-Karpyn, line 40:
```suggestion
  return {
```

Ihor-Karpyn (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    },
+    divide(a, b) {
+      return a / b;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+  };
```
Ihor-Karpyn, line 64: Better to leave empty line between methods

Ihor-Karpyn (approved):

PR #782, opened 2022-07-12
pocket calculator solution

Vova-Balyshev:


budanov20:
all right

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+  calculator.divide = function(a, b) {
+    return a / b;
+  };
+
+  calculator.reset = function() {
+    calculator.result = 0;
+
+    return this;
+  };
+
```
ivan-kadykalo, line 65: It is a possible solution, but why you don't write all these methods inside your object? It will allow you to use `this`.

ivan-kadykalo (changes requested):

src/makeCalculator.js
Vova-Balyshev: I made like this because Ihor asked to make that way, i forgot the correct name of this, but the point is that we take properties only inside function. @ivan-kadykalo

PR #797, opened 2022-07-12
Make calculator task solution

juliiaap:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
+
+      return calculator.result;
+    },
+
+    subtract(number) {
+      calculator.result -= number;
```
alexandra-protyanova, line 50:
```suggestion
      result -= number;
```

alexandra-protyanova: you can call it just result

alexandra-protyanova (approved):

PR #798, opened 2022-07-12
Solution

LovelyDrug:


rodiontseva (approved):
Great and easy to understand (without using this)!

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+  const calculator = {
+    result: 0,
+    operate(operation, number) {
+      operation(number);
+
+      return calculator;
+    },
+    add(n) {
+      calculator.result += n;
+    },
```
ivan-kadykalo, line 49: It would be better to return result and save him in operate method.
```suggestion
    add(a, b) {
      return a + b;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(operation, number) {
+      operation(number);
```
ivan-kadykalo, line 43:
```suggestion
      this.result = operation(this.result, number);
```

ivan-kadykalo (changes requested):

Ihor-Karpyn (approved):
Good job

PR #809, opened 2022-07-18
solution

dima-korovai:


src/makeCalculator.js
```diff
@@ -37,7 +37,50 @@
...
+
+          return this;
+
+        case calculator.divide:
+          this.result /= number;
+
+          return this;
+      }
+    },
+  };
```
valerii-derkach, line 81: move logic with actions inside actions add / divide /  subtract / multiply / 
```suggestion
    operate(operator, number) {
      this.result = operator(this.result, number);

      return this;
    },
```

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #826, opened 2022-08-03
Task_advanced-calculator

MarkShapka:


src/makeCalculator.js
```diff
@@ -38,6 +38,48 @@
...
+        this.subtract(number);
+      }
+
+      if (callback === this.multiply) {
+        this.multiply(number);
+      }
+
+      if (callback === this.divide) {
+        this.divide(number);
+      }
```
alexandra-protyanova, line 75: just use callback and pass number there, logic is the same for everything

alexandra-protyanova (changes requested):

PR #829, opened 2022-08-06
solution

Kulchevych:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(number) {
+      return this.result + number;
+    },
```
tiserett, line 45:
```suggestion
    add(number) {
      return this.result + number;
    },
```
you can get rid of all these returns, and adjust result's value right here

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation.call(this, number);
```
tiserett, line 66:
```suggestion
      operation.call(this, number);
```
now you can simply execute function, without setting it's return to result

tiserett (approved):
everything  is good, but if you want, you can simplify your code, as I showed

tiserett (approved):
good job ;)

danylo-onishchuk (approved):
nice )

PR #833, opened 2022-08-09
Solution

Dmytro-Matsehora:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
tiserett, line 40:
```suggestion
  const calculator = {
```

no need to create new variable for storing object, you can immediately return it,

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    operate(callback, n) {
+      callback.call(this, n);
+
+      return this;
+    },
+  };
+
+  return calculator;
```
tiserett, line 72:
```suggestion
  return calculator;
```
we don't need it anymore

tiserett (changes requested):
just a few small fixes required

tiserett (approved):
good job!

danylo-onishchuk (approved):
good job!

PR #838, opened 2022-08-10
Solution

VolodymyrChuyko:


src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: function(initValue, argValue) {
```
plemiannik, line 43: why it in such form here and in other cases not?

plemiannik (approved):
Very interesting and usefull solution 4 me, cos i thinked, that i must use get if i want 2 use functions without () 👍

src/makeCalculator.js
VolodymyrChuyko: you're right, it could be shortened

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(initValue, argValue) {
+      return initValue + argValue;
+    },
```
AlinaMatuschak, line 45: you can use result from calculator

```suggestion
    add(argValue) {
      return calculator.result + argValue;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    divide(initValue, argValue) {
+      return initValue / argValue;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(operation, argument) {
+      this.result = operation(this.result, argument);
```
AlinaMatuschak, line 61: 😅

AlinaMatuschak (approved):
Interesting solution 🤓

src/makeCalculator.js
VolodymyrChuyko: Thanks, I've simplified my code, using your advice

PR #840, opened 2022-08-12
Solution

andrii-dmitruk:


kostya-kosyuk (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(augend, addend) {
+      return augend + addend;
+    },
```
AlinaMatuschak, line 57: you can use result from your calculator

```suggestion
    add(addend) {
      return calculator.result + addend;
    },
```

AlinaMatuschak (approved):
Interesting solution 🤓

src/makeCalculator.js
andrii-dmitruk: I tried to do it like this:
```
add(addend) {
      this.result += addend;
    },
```
But it didn`t work, so i thought that it would be a good idea to put result like an argument of function. I forgot that _object._  could be used instead of _this._

AlinaMatuschak: Need more read about this )
But `add` is a callback, the function doesn’t have this:

PR #842, opened 2022-08-13
done

adrqq:


oksana-logos-frontend (approved):
Good job)) Have you read the checklist?
You could write method like:
 add(value) {
      this.result += value;
    }, 
that would be better in my opinion.

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate: (operation, value) => {
+      calculator.result = operation(value);
+
+      return calculator;
+    },
```
AlinaMatuschak, line 71: You can use this on operate and reset
When you use constructors, classes, or something like that this looks to service which call your method
calculator.reset() -> this on reset is a calculator
But on run callback, the function doesn’t have this:
add() -> this isn’t exit here

AlinaMatuschak (changes requested):
Just small issues 👇

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      callback.call(this, number);
```
AlinaMatuschak, line 44: 🔥

AlinaMatuschak (approved):
👏

PR #844, opened 2022-08-14
Solution

plemiannik:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    get add() {
+      return (a, b) => a + b;
+    },
```
VolodymyrChuyko, line 45:
```suggestion
    add(a, b) {
      return a + b;
    },
```
you can use functions instead of getters

VolodymyrChuyko (changes requested):

src/makeCalculator.js
plemiannik: i already understand it looking at ur solution, but why i must use functions instead of getters?

VolodymyrChuyko (approved):
nice!

dellachaise (approved):
Well done 👍

PR #861, opened 2022-08-15
task solution code

rboiar:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const operations = {
```
Jamal-Amoian, line 40: What  does you mean, when you called an object as a 'operations', I think, it's better, if you called this one as a 'calculator'.

Jamal-Amoian (changes requested):

src/makeCalculator.js
rboiar: > What does you mean, when you called an object as a 'operations', I think, it's better, if you called this one as a 'calculator'.

In the name of the function, it is quite clear that this is a calculator

So in an object, I use the name "operations", because what happens in the function is exactly operations or "actions"

Add, subtract, and so on... ;)

Jamal-Amoian (approved):
Great job!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, operand) {
+      operation.call(this, operand);
```
AlinaMatuschak, line 50: 🔥

AlinaMatuschak (approved):
🚀🚀🚀

PR #862, opened 2022-08-15
Advanced calculator

Desertiks:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+    subtract() {
+      return 'subtract';
+    },
+    divide() {
+      return 'divide';
+    },
+    multiply() {
+      return 'multiply';
+    },
```
yungdena, line 53:
```suggestion
    },    add(number) {
      this.result += number;
    },

    subtract(number) {
      this.result -= number;
    },

    multiply(number) {
      this.result *= number;
    },

    divide(number) {
      this.result /= number;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      return 'add';
+    },
+    subtract() {
+      return 'subtract';
+    },
+    divide() {
+      return 'divide';
+    },
+    multiply() {
+      return 'multiply';
```
yungdena, line 52: change the result, instead of returning a string

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+    operate(operator, number) {
+      if (operator() === 'add') {
+        this.result += number;
+      }
```
yungdena, line 57: you shouldn't use if statements,  use callback

yungdena (changes requested):

src/makeCalculator.js
Desertiks: Ok, i fixed this. I hope I understood the task correctly this time. 😄

yungdena (approved):
good job

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+    add(number) {
+      calculator.result += number;
+    },
+    subtract(number) {
+      calculator.result -= number;
+    },
+    divide(number) {
```
dellachaise, line 48: Add page breaks.
```suggestion
  const calculator = {
    result: 0,
    
    add(number) {
      calculator.result += number;
    },
    
    subtract(number) {
      calculator.result -= number;
    },
    
    divide(number) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      calculator.result /= number;
+    },
+    multiply(number) {
+      calculator.result *= number;
```
dellachaise, line 52: In all methods replace `calculator` with `this. If you would change `calculator` to another  name, this way you would need to change only one line, not each method.
```suggestion
      this.result *= number;
```

dellachaise (changes requested):
Great job 👍 

Just add a few fixes according to comments 👇

src/makeCalculator.js
Desertiks: I dont know whyI can't use 'this' directly in this methods. I needed to pass an argument 'this' before that.

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
dellachaise, line 40: Try to rewrite this way
```suggestion
  return {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    result: 0,
+
+    add(number, obj) {
+      obj.result += number;
+    },
```
dellachaise, line 45: And this
```suggestion
    add(number) {
      this.result += number;
    },
```

dellachaise (changes requested):
Try to rewrite, if you would have some trouble - write to qna chat and come to qna today at 17:00.

dellachaise (approved):
Well done 👍

PR #866, opened 2022-08-16
solution

OleksiiPustovit:


anyakrainiuk (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(value) {
+      this.result += value;
+    },
+    subtract(value) {
+      this.result -= value;
```
dellachaise, line 46: Better add page breaks all over the code, like here 👇 
```suggestion
  return {
    result: 0,
    
    add(value) {
      this.result += value;
    },
    
    subtract(value) {
      this.result -= value;
```

dellachaise (approved):
Great job 👍 

See more in comments 👇

PR #868, opened 2022-08-16
Solution

DimaKobzar7:


Alzay007 (approved):
great job 👍

DimaKobzar7:
Thanks!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
```
AlinaMatuschak, line 47: Here you can use this too

But add is a callback, the function doesn’t have this:

```suggestion
    reset() {
      this.result = 0;

      return this;
    },
```

AlinaMatuschak (approved):
Good work 👍

src/makeCalculator.js
Alzay007: thanks for your advice)

PR #874, opened 2022-08-16
solution

zinchenko-anastasiia:


nikolay-li (approved):
good job ;-)

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      calculator.result += value;
+    },
+
+    subtract(value) {
+      calculator.result -= value;
```
dellachaise, line 48: Try to rewrite all methods with this. Because if you change object name in line 40, you should change it in all methods. Using this - no such a problems.

```suggestion
      this.result += value;
    },

    subtract(value) {
      this.result -= value;
```

dellachaise (changes requested):
Great job 👍 

But please add changes according to comment 👇

dellachaise (approved):
Well done 👍

PR #875, opened 2022-08-16
calc

oksana-kyryienko:


src/makeCalculator.js
```diff
@@ -1,43 +1,51 @@
...
+      operation.call(this, value);
+
+      return this;
+    },
+
+    result: 0,
+
+    clear() {
+      return this.result;
+    },
```
VasylZinchenko, line 38: probably it is unnecessary

VasylZinchenko (approved):
👍

src/makeCalculator.js
```diff
@@ -1,43 +1,51 @@
...
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      this.result /= value;
+    },
+
+    operate(operation, value) {
+      operation.call(this, value);
```
AlinaMatuschak, line 29: 🔥

src/makeCalculator.js
```diff
@@ -1,43 +1,51 @@
...
+
+      return this;
+    },
+
+    result: 0,
```
AlinaMatuschak, line 34: it should be on the top.
Better write simple values first, next getters and setters, and next methods. It will be easier to find what you need if have one structure

AlinaMatuschak (approved):
Good work 👏

PR #878, opened 2022-08-16
Solution

Roman-Artiukh:


src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
...
+    result: 0,
+
+    operate(operation, operand) {
+      operation(operand);
+
+      return this;
+    },
+
+    add(value) {
+      calculator.result += value;
```
roma-shafihullin, line 52:
```suggestion
      this.result += value;
```

roma-shafihullin (approved):
good work!

src/makeCalculator.js
roma-shafihullin: and further through the code

AlinaMatuschak: Agree with @roma-shafihullin here )

But add is a callback, the function doesn’t have this:

```suggestion
    reset() {
      this.result = 0;

      return this;
    },
```

AlinaMatuschak (approved):

PR #881, opened 2022-08-16
Solution

max-kravchenko:


vladgusakovskiy (approved):
Good job!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    multiply(number) {
+      calculator.result *= number;
+    },
+
+    divide(number) {
+      calculator.result /= number;
+    },
+
+    reset() {
+      calculator.result = 0;
```
AlinaMatuschak, line 60: You can use this here

But `add` is a callback, the function doesn’t have this:

AlinaMatuschak (approved):
Good work 👏

PR #885, opened 2022-08-17
calculator

Reptiloid044:
Hi!

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, value) {
+      callback(value);
+
+      return this;
+    },
+
+    add(value) {
+      calculator.result += value;
+    },
+
+    subtract(value) {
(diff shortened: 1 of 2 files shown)


oksana-logos-frontend (approved):
Great job!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    operate(callback, value) {
+      callback(value);
+
+      return this;
+    },
+
+    add(value) {
+      calculator.result += value;
+    },
```
dellachaise, line 51: try to rewrite this way. Because if you change variable name, you should add a lot of changes.
With this solution better in 40 line return object without creating variable.
```suggestion
    add(value) {
      this.result += value;
    },
```

dellachaise (changes requested):
Great start 👍 

Please add changes according to comment 👇

Reptiloid044:
Hi all!

Thank you for the review and comment, I had fixed it according to your request, please check, it seems that is right way=)

Have a nice day!

dellachaise (approved):
Well done 👍

PR #894, opened 2022-08-26
Solution

kate-matviiv:


src/makeCalculator.js
```diff
@@ -36,8 +36,35 @@
...
  * @return {object}
  */
+
 function makeCalculator() {
   // write code here
-}
+  const calculator = {
+    result: 0,
+    add: function(number) {
+      calculator.result += number;
```
alexandra-protyanova, line 45: it is callback it should receive 2 parameters and return calculation

alexandra-protyanova (changes requested):

ArturErmolenko (approved):

PR #899, opened 2022-08-31
Solution

AMahalias:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+  const calculator = {
+    result: 0,
+    add(a) {
+      calculator.result += a;
+
+      return calculator.result;
+    },
+    subtract(a) {
+      calculator.result -= a;
+    },
```
stanislav-topikha, line 49: Add default values and blank lines
why in one case you're returning smth and in other dont?
let's be consistent
```suggestion
  const calculator = {
    result: 0,
    
    add(a = 0) {
      calculator.result += a;

      return calculator.result;
    },
    
    subtract(a = 0) {
      calculator.result -= a;
    },
```

stanislav-topikha (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      return calculator.result;
+    },
+
+    subtract(a = 0) {
+      calculator.result -= a;
+
+      return calculator.result;
+    },
+
+    multiply(a = 0) {
```
stanislav-topikha, line 55: so we gonna set result value to zero by default?

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      calculator.result *= a;
+
+      return calculator.result;
+    },
+
+    divide(a = 0) {
```
stanislav-topikha, line 61: is it possible to divide by zero?

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      return calculator.result;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(operation, value = 0) {
```
stanislav-topikha, line 73:
```suggestion
    operate(operation, value) {
```

stanislav-topikha (changes requested):
comments ⬇️

liliia-livandovskaya (approved):
GJ 👍

PR #900, opened 2022-08-31
Solution

illia-kots:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      calculator.result *= number;
+    },
+    divide: function(number) {
+      calculator.result /= number;
+    },
+    reset: function(number) {
+      calculator.result = 0;
+
+      return this;
+    },
```
SanyaBratashchuk, line 63: use methods for example divide(number) {...some code}

SanyaBratashchuk (changes requested):

src/makeCalculator.js
illia-kots: Done

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      calculator.result += number;
+    },
+    subtract(number) {
+      calculator.result -= number;
+    },
+    multiply(number) {
+      calculator.result *= number;
+    },
+    divide(number) {
+      calculator.result /= number;
```
SanyaBratashchuk, line 57: do not interact with result in this methods return result of calculation and assign it in operate method

SanyaBratashchuk (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    add(a, b) {
+      return a + b;
+    },
+    subtract(a, b) {
+      return a - b;
+    },
+    multiply(a, b) {
+      return a * b;
+    },
+    divide(a, br) {
```
SanyaBratashchuk, line 56:
```suggestion
    divide(a, b) {
```

SanyaBratashchuk (approved):

PR #904, opened 2022-09-04
solution

illia-kyselov:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      calculator.result -= number;
+    },
+    divide: (number) => {
+      calculator.result /= number;
+    },
+    multiply: (number) => {
+      calculator.result *= number;
+    },
+    operate: (callback, number) => {
+      callback(number);
```
alexandra-protyanova, line 55: pass result and incoming number here, don't manipulate with result in callbacks

alexandra-protyanova (changes requested):

src/makeCalculator.js
illia-kyselov: I don't understand what is problem?

alexandra-protyanova: pass `this.result` as an argument to callback, in callbacks just return result

alexandra-protyanova (changes requested):

SanyaBratashchuk (approved):

PR #908, opened 2022-09-08
done

dmitriyn3679:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(action, num) {
+      action(num, this);
```
alexandra-protyanova, line 50:
```suggestion
      action(num, this.result);
```

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+      return this;
+    },
+
+    add(num, currentObj) {
+      currentObj.result += num;
```
alexandra-protyanova, line 56: and here u just can get result

alexandra-protyanova (changes requested):

src/makeCalculator.js
dmitriyn3679: subtract(num, result) {
      result -= num;
    },
    
but i reassign param

dmitriyn3679: > function should return calculation

subtract(num, result) {
result -= num;
},

but i reassign param

marina-tilniak: you return from operator, so it's ok

marina-tilniak (approved):

PR #909, opened 2022-09-09
Task solved

sergey-vorobei:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      callback(number);
```
alexandra-protyanova, line 44: pass result as an argument to callback

src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+    add(number) {
+      calculator.result += number;
+    },
+
+    subtract(number) {
+      calculator.result -= number;
+    },
 
+    divide(number) {
+      calculator.result /= number;
```
alexandra-protyanova, line 58: don't interact with the object's result, just return calculation and set it in operate function

alexandra-protyanova (changes requested):

src/makeCalculator.js
sergey-vorobei: could you help me, how i can do it? I little understand all how work it

alexandra-protyanova (changes requested):
ask your question in fe_chat, it is easier to help u there

src/makeCalculator.js
```diff
@@ -37,7 +39,38 @@
...
+    divide(a, b) {
+      return a / b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
 
+    reset() {
+      calculator.result = 0;
```
SanyaBratashchuk, line 62:
```suggestion
      this.result = 0;
```

SanyaBratashchuk (approved):

PR #919, opened 2022-09-14
calculator solution

TheVovchik:


imaksweb (approved):
Great job 🚀

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+      return calculatedValue;
+    },
+    multiply(startValue, valueFromUser) {
+      const calculatedValue = startValue * valueFromUser;
+
+      return calculatedValue;
+    },
+    operate(callback, valueFromUser) {
+      this.result = callback(this.result, valueFromUser);
```
Anna-Tkachenko, line 67: will be better to create more understandable name for callback function
for now it's not clear, what result we expect to receive from this function, and how this function can works

Anna-Tkachenko (approved):

PR #921, opened 2022-09-15
calculator solution

vmamchur:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
```
grafit-off, line 44: Some students call the argument of "operate" just `callback`, it's not really clear to me, but `operation` it's really clear! GJ 🔥

grafit-off (approved):
GJ 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(initial, toOperate) {
+      return initial + toOperate;
+    },
```
Anna-Tkachenko, line 57: each function just accept 2 arguments, maybe it's not so clear why one of them is initial, and one of them is toOperate
what does it mean? let's simplify variable names, it's just 2 arguments, some 2 values that we need to proceed

Anna-Tkachenko (approved):

PR #929, opened 2022-09-16
Solution

imaksweb:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      callback.call(this, value);
```
TheVovchik, line 66: as I understand call() -  It can be used to invoke (call) a method with an owner object as an argument (parameter).
and in this case you transfer whole object, not a single parametr, not sure is this a god practice here.

With call(), an object can use a method belonging to another object.
Like in case with call we can create even a side object, and take operate method from there. 

will appreciate if you explicate why you use it here.

TheVovchik (approved):

src/makeCalculator.js
imaksweb: It means that I invoke my `callback` function in context of my object, because in this case I loose my context, and without binding my `this` = window global object. If I invoke `callback` without binding: 

```js
callback(value); // for example I invoke add method

add(value) {
  this.result += value; // without binding it looks like window.result += value
}
// I lost my context now it's window global object not my object
```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      if (value === 0) {
+        throw new Error('Dividing by zero!');
+      }
```
TheVovchik, line 58: 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(value) {
+      this.result += value;
+    },
```
Anna-Tkachenko, line 45: each method function shouldn't know anything about the result property
it should accept 2 arguments, and return some result depending on own functionality

Anna-Tkachenko (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(prevValue, value) {
+      return prevValue + value;
+    },
```
Anna-Tkachenko, line 45: both of arguments should have the same meaning
why one of them is prev value, and other one is value?
let's think more about names for variables

Anna-Tkachenko (approved):

PR #934, opened 2022-09-17
Solution of Advanced calculator

Dikey945:


imaksweb (approved):
GJ 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  // object created
+  const calculator = {
+    result: 0,
+
+    // created callback operator methods
+    add: function(prevValue, number) {
```
Anna-Tkachenko, line 45: check point 4 in checklist

Anna-Tkachenko (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
 function makeCalculator() {
-  // write code here
+  // object created
+  const calculator = {
+    result: 0,
+
+    // created callback operator methods
+    add(prevValue, number) {
+      return prevValue + number;
+    },
```
Anna-Tkachenko, line 47: why one number is prev value? and other one is simple number?
let's think more about variable names, they should have the same meaning

Anna-Tkachenko (approved):
Good job!

PR #937, opened 2022-09-18
solution

nikitaovch99:


volodymyrklymkiv (approved):
Great

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+  const calculator = {
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(value, number) {
```
Anna-Tkachenko, line 49: what's the difference between value and number? think about more understandable variable names for 2 arguments in operation function

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    divide(value, number) {
+      return value / number;
+    },
+
+    multiply(value, number) {
+      return value * number;
+    },
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
```
Anna-Tkachenko, line 66: same with callback variable name, it's not clear how this function works according to the name

Anna-Tkachenko (approved):

PR #941, opened 2022-09-18
Solution

Rina07:


Oleksandra-Shevchenko (approved):
Good job 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num) {
+      calculator.result += num;
+
+      return calculator;
+    },
```
OleksiiBlezniuk, line 47: This method shouldn't have access to `result`, assigning to `result` should be in `operate` method
Let's pass two parameters here and return result of operation. Then in `operate` we'll assign result of operation to `result`

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+    divide(num) {
+      calculator.result /= num;
+
+      return calculator;
+    },
+
+    operate(callback, num) {
+      return callback(num);
+    },
```
OleksiiBlezniuk, line 69: assign operation result to `result` property here and return `this`

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+    reset() {
+      calculator.result = 0;
```
OleksiiBlezniuk, line 72: use `this`

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
GJ

PR #947, opened 2022-09-18
solution

myroslav-diiak:


jojnajbude (approved):

src/makeCalculator.js
```diff
@@ -38,6 +38,47 @@
...
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      calculator.result += value;
+
+      return calculator;
+    },
```
OleksiiBlezniuk, line 48: This method should just calculate and return result of calculation
Let's assign value to result in operate method

OleksiiBlezniuk (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,47 @@
...
+
+    divide(value) {
+      calculator.result /= value;
+
+      return calculator;
+    },
+
+    reset() {
+      calculator.result = 0;
```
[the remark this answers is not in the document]
OleksiiBlezniuk: use `this` here

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    },
+
+    reset(value) {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, value) {
+      calculator.result = callback(value);
```
OleksiiBlezniuk, line 67: use `this` here

OleksiiBlezniuk (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      return calculator.result + value;
+    },
```
OleksiiBlezniuk, line 46: This is method that should add two values and return result of calculation
It shouldn't have access to parent object and it's properties

OleksiiBlezniuk (changes requested):

PR #948, opened 2022-09-19
Solution

Apollonia5:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(value) {
+      calculator.result += value;
+    },
```
imaksweb, line 44: Methods `add`, `subtract`, `multiply` and `divide` shouldn't know anything about the `result` property, the should accept 2 arguments and return some result depending on their own functionality 🙃

imaksweb (changes requested):

silchyn (approved):
Cool!

PR #949, opened 2022-09-19
solution was added

Maksym845:


Yeliseiev (approved):
Good job.  🚀

silchyn (changes requested):
Does "res" instead of "result" makes big difference?

silchyn (approved):
Cool!

PR #956, opened 2022-09-19
implemented calculator

a-nahaitsev:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    subtract(a, b) {
+      return a - b;
+    },
+    multiply(a, b) {
+      return a * b;
+    },
+    divide(a, b) {
+      return a / b;
+    },
+  };
```
anastasiia-nahaiska, line 64: Maybe it'll be better to write "return calculator", but I'm not sure
image

anastasiia-nahaiska (approved):
Great!

Anna-Tkachenko (approved):
Good job!

PR #957, opened 2022-09-19
Function makeCalculator was created

mykyta-krainik:


src/makeCalculator.js
```diff
@@ -38,6 +38,37 @@
  */
 function makeCalculator() {
   // write code here
```
liliia-livandovskaya, line 40: remove comments

liliia-livandovskaya (approved):
GJ 🔥

PR #962, opened 2022-09-19
calculator

Arsen-Shpak:


src/makeCalculator.js
```diff
@@ -38,6 +38,33 @@
...
+      return this;
+    },
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+  };
+
+  return obj;
```
Andrew-Skoryk, line 67: 1. [NAMING] - Use proper names for variables. If function was called makeFruit, it should return fruit.
I would recommend changing it to calculator=)

Andrew-Skoryk (approved):

silchyn (changes requested):
Check the checklist

silchyn (approved):
Cool!

PR #971, opened 2022-09-20
solution

igor-1987:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
```
dmytristepanov, line 56: I think you need to use more understandable names of parameters

dmytristepanov (approved):
Well done!
Just left a hint

src/makeCalculator.js
igor-1987: > I think you need to use more understandable names of parameters

Yeah, you are right. Thank you!

OleksiiBlezniuk (approved):
GJ!

PR #973, opened 2022-09-20
Solution

dmytristepanov:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(firstValue, secondValue) {
+      const resultAdd = firstValue + secondValue;
+
+      return resultAdd;
```
igor-1987, line 46: Well done! 
As option, you can reduce your code a little bit on rows similar to these.

igor-1987 (approved):

src/makeCalculator.js
silchyn, line 46:
```suggestion
      return firstValue + secondValue;
```

silchyn (changes requested):

silchyn (approved):
Cool!

PR #976, opened 2022-09-21
solution

Naveed-Shoukat:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    operate(operation, number) {
+      operation.call(this, number);
+
+      return this;
+    },
+
+  };
+
+  return calculator;
 }
```
kamal-tufekcic, line 74:
```suggestion
  return {
    result: 0,

    add(number) {
      this.result += number;
    },

    subtract(number) {
      this.result -= number;
    },

    multiply(number) {
      this.result *= number;
    },

    divide(number) {
      this.result /= number;
    },

    reset() {
      this.result = 0;

      return this;
    },

    operate(operation, number) {
      operation.call(this, number);

      return this;
    },
  };
}
```

Just a suggestion, you can also return immediately since you don't modify the object after creating it

kamal-tufekcic (approved):

ihor-jpeg (approved):

PR #978, opened 2022-09-21
Solution

AnastasiaTitarenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    subtract(number) {
+      obj.result -= number;
+    },
+    multiply(number) {
+      obj.result *= number;
+    },
+    divide(number) {
+      obj.result /= number;
+    },
+    reset() {
```
Bohdan-Maksymiuk, line 59: add more free space between methods

Bohdan-Maksymiuk (changes requested):
try to use `this` instead of direct access through `obj`

Bohdan-Maksymiuk (approved):

PR #981, opened 2022-09-22
solution

oleksii-bieliaiev:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+      calculator.result -= num;
+    },
+
+    multiply(num) {
+      calculator.result *= num;
+    },
+
+    divide(num) {
+      calculator.result /= num;
+    },
```
SanyaBratashchuk, line 57: do not change result in this methods, return only value and change result in operate method

SanyaBratashchuk (changes requested):

src/makeCalculator.js
oleksii-bieliaiev: if I understand correctly, something like this

qoa11a (approved):

PR #988, opened 2022-09-25
solution

bradleymatjie:


src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+      return currentResult - num;
+    },
+
+    multiply(currentResult, num) {
+      return currentResult * num;
+    },
+
+    divide(currentResult, num) {
+      return currentResult / num;
+    },
```
kamal-tufekcic, line 64: These should take just 1 parameter and modify the result

kamal-tufekcic (approved):

VinnieJ-2k20 (approved):

PR #993, opened 2022-09-27
Solution

vaadyak:


src/makeCalculator.js
```diff
@@ -38,6 +38,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      callback.call(this, value);
```
MaxManis, line 67: You also can do it without using of `call` but okey)

MaxManis (approved):
GJ

PR #1000, opened 2022-09-30
Solution1

JohnRust12:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(res, number) {
+      return res + number;
+    },
```
ArturErmolenko, line 45: you need to increase calculator result like this
```
add(value) {
      return calculator.result + value;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
```
ArturErmolenko, line 68:
```suggestion
  operate(callback, value) {
      this.result = callback(value);

      return this;
    },
```

ArturErmolenko: use proper names for variables - if you want to name variable result, pick a better one :)

ArturErmolenko (changes requested):

SanyaBratashchuk (approved):

PR #1001, opened 2022-10-01
vol_1

RudBeee:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      calculator.result /= number;
+    },
+    multiply: (number) => {
+      calculator.result *= number;
+    },
+    operate: (callback, number) => {
+      callback(number);
+
+      return calculator;
+    },
```
ivanchuk-yurii, line 58: return value from functions and assign `calculator.result` in `operate`

ivanchuk-yurii (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: (number) => {
+      return calculator.result + number;
```
AlisaVlasova, line 43:
```suggestion
      return this.result + number;
```

AlisaVlasova (changes requested):

src/makeCalculator.js
RudBeee: Якщо змінюю на this,  то тести не проходить(

loralevitska (approved):
for `this` to work, you need to return an object and not a variable that is an object. Try experimenting with `this` one for better understanding

PR #1006, opened 2022-10-06
Develop

metinbicaksiz:


src/makeCalculator.js
```diff
@@ -38,6 +38,41 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    add(a) {
+      calculator.result = calculator.result + a;
```
amitpatiljc, line 44: can be written in a shorter way

src/makeCalculator.js
```diff
@@ -38,6 +38,41 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    add(a) {
```
amitpatiljc, line 43: let's use relevant variable name

amitpatiljc (changes requested):

src/makeCalculator.js
amitpatiljc: may be `number`

src/makeCalculator.js
```diff
@@ -38,6 +38,41 @@
...
+      calculator.result += a;
+
+      return this;
+    },
+    subtract(a) {
+      calculator.result -= a;
+
```
amitpatiljc, line 50: variable name can be like `number` ?

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job!

PR #1009, opened 2022-10-08
Solution

uliana-kudriavtseva:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+    result: 0,
+
+    add(value = 0) {
+      this.result += value;
+    },
+    subtract(value = 0) {
+      this.result -= value;
+    },
+    multiply(value = 0) {
+      this.result *= value;
```
SanyaBratashchuk, line 50: change `result` only in `operate` and `reset` method

SanyaBratashchuk (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      return this.result + value;
+    },
```
marina-tilniak, line 45: would be good add empty line after each method

marina-tilniak (approved):

PR #1015, opened 2022-10-13
commit

devWizHub:


src/makeCalculator.js
```diff
@@ -38,6 +38,47 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
```
jstmpelowycz, line 45: let's use `this` instead of `calculator`, that would be more consistent
you'll hit into the issues with context loss, so here is the relevant article on this problem
https://www.freecodecamp.org/news/what-to-do-when-this-loses-context-f09664af076f/

jstmpelowycz (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,53 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+  };
+
+  function reset() {
+    this.result *= 0;
```
amitpatiljc, line 46:
```suggestion
    this.result = 0;
```

src/makeCalculator.js
devWizHub: Implemented.

src/makeCalculator.js
devWizHub: changed

amitpatiljc (approved):
Good job 👍

PR #1016, opened 2022-10-14
Solution

Julia2063:


Bohdan-Maksymiuk (approved):
as a challenge - you may try to rewrite with using `this` instead of calculator in methods

PR #1019, opened 2022-10-16
Done

ruslik87:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    multiply(value) {
+      return calculator.result * value;
+    },
+
+    divide(value) {
+      return calculator.result / value;
+    },
+
+    operate(operation, value) {
+      this.result = operation(value);
```
alexandra-protyanova, line 58: let us pass here `this.result` as well

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    add(value) {
+      return calculator.result + value;
```
alexandra-protyanova, line 42: and let us not touch object here, just return value and set it in `operate`

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1023, opened 2022-10-24
Solution

Illia-Skladnik:


src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+      calculator.result -= num;
+    },
+
+    multiply(num) {
+      calculator.result *= num;
+    },
+
+    divide(num) {
+      calculator.result /= num;
+    },
```
illiarerun, line 64: can we divide by `0`?

illiarerun (changes requested):

src/makeCalculator.js
```diff
@@ -36,8 +36,47 @@
...
+
+    divide(num1, num2) {
+      switch (num1) {
+        case 0:
+          throw new Error('cannot be divided by zero');
+
+        default:
+          return num2 / num1;
+      }
```
illiarerun, line 69: switch is an overkill, here you can just use `if`

illiarerun (approved):

PR #1024, opened 2022-10-24
solution

MaxFedak:


src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
```
denys-danyliuk, line 48:
```suggestion
    reset() {
      this.result = 0;

      return this;
    },
```

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+
+    add(number) {
+      calculator.result += number;
+    },
```
denys-danyliuk, line 58: Make it like a pure function.
For example

```
add(num1, num2) {
	return num1 + num2;
}

File: src/makeCalculator.js [hunk 3]
```diff
@@ -38,6 +38,39 @@
...
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
```
denys-danyliuk, line 54: save your result here
```
this.result = callback ...

**Review by denys-danyliuk — CHANGES_REQUESTED** [review]

File: src/makeCalculator.js [hunk 4]
```diff
@@ -38,6 +38,41 @@
...
+
+    multiply(number1, number2) {
+      return number2 * number1;
+    },
+
+    divide(number1, number2) {
+      if (number2 !== 0) {
+        return number2 / number1;
+      }
```
illiarerun, line 71: better throw some error if we pass 0

src/makeCalculator.js
```diff
@@ -38,6 +38,41 @@
...
+    subtract(number1, number2) {
+      return number2 - number1;
+    },
+
+    multiply(number1, number2) {
+      return number2 * number1;
+    },
+
+    divide(number1, number2) {
+      if (number2 !== 0) {
```
illiarerun, line 69: you should check the **first number** for 0 as its a denominator and it can't be 0

illiarerun (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,43 @@
...
+
+    multiply(number1, number2) {
+      return number2 * number1;
+    },
+
+    divide(number1, number2) {
+      if (number1 !== 0) {
+        return number2 / number1;
+      } else {
+        throw new Error(`U can't devide by 0, mate`);
```
illiarerun, line 72: )
```suggestion
        throw new Error(`U can't divide by 0, mate`);
```

illiarerun (approved):

src/makeCalculator.js
MaxFedak: oh)))

PR #1026, opened 2022-10-26
Calculator was created

iryna-lelyk:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num1, num2) {
```
denys-danyliuk, line 43:
```suggestion
    add(currentValue, num2) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
```
denys-danyliuk, line 40: you need to create object and then return it, so we can call methods like `calculator.add()`
Знімок екрана 2022-10-26 о 15 49 08

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #1029, opened 2022-10-26
Solution

andriy-shymkiv:


src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+      callback(number);
+
+      return this;
+    },
+
+    add(num) {
+      calculator.result += num;
+
+      return this;
+    },
```
Svitlana-Yudina, line 59: You don't need return here, I think)

Svitlana-Yudina (approved):
It's a good solution as for me) 🔥

src/makeCalculator.js
Andriy-Shymkiv: I will try to remove return, but i'm not sure if it will work properly)

illiarerun: You don't call these methods explicitly, so indeed you don't need it

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
```
illiarerun, line 44: right now we can execute this code:
image
i guess it should work a bit other way:
- we should change object that we called `operate` **on**,
- operation shouldn't be dependant on exact object that you call it from
- you can somehow pass needed object from operate to this actual `operation`, try `this`

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+      calculator.result += num;
+
+      return this;
+    },
+
+    subtract(num) {
+      calculator.result -= num;
+
+      return this;
+    },
```
illiarerun, line 65: make this functions like a [pure ones](https://www.geeksforgeeks.org/pure-functions-in-javascript/), pass your result as first param, and number as a second param. do it so your operations won't be dependant on previous result

illiarerun (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+
+    multiply(num1, num2) {
+      return num1 * num2;
+    },
+
+    divide(num1, num2) {
+      if (num1 === 0) {
+        throw Error(`We can't divide by zero`);
+      }
```
illiarerun, line 70: we should check divisor (дільник), not the divident (ділене)

illiarerun (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      return num1 / num2;
+    },
+  };
+
+  return calculator;
 }
 
+const calculatorre = makeCalculator();
+
+console.log(calculatorre.operate(calculatorre.add, 10).operate(calculatorre.divide, 0));
```
mykola-koval, line 81: remove logs

mykola-koval (changes requested):

illiarerun (approved):

PR #1033, opened 2022-10-26
Solution

referee1988:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
```
masha-girya, line 44: i think, you shoult add 'this' instead of 'calculator' here

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+      return this;
+    },
+
+    add(operand) {
+      calculator.result += operand;
```
masha-girya, line 50: nice, i have 'this.result' (it did not work)))) 'calculator.result' is good idea, but what if we need to change object's name?

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    subtract(operand) {
+      calculator.result -= operand;
+    },
+
+    divide(operand) {
+      calculator.result /= operand;
+    },
```
masha-girya, line 59: what if there will be zero?)) we can't divide by zero (it was mentor's advice for me)

referee1988: I agree. Added check for 0.

src/makeCalculator.js
referee1988: The object calculator is wrapped in the makeCalculator() function. The purpose of the function is to use it "as is", without changing anything in it.
It’s a pity that JS doesn’t have encapsulation as such and access modifiers that could be used to forcibly close access to it.

Assigning any name to the function return (const calculator = makeCalculator(); const myCalculator = makeCalculator(); const someCalculator = makeCalculator(); ...) will not break the function.

src/makeCalculator.js
denys-danyliuk: agree

src/makeCalculator.js
denys-danyliuk: it won't, but you surely need to use this in your object methods

src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
+      return this;
+    },
+
+    add(operand) {
+      calculator.result += operand;
+    },
+
+    subtract(operand) {
+      calculator.result -= operand;
+    },
```
denys-danyliuk, line 55: Make it like a pure function.
For example
```
add(num1, num2) {
	return num1 + num2;
}
```

src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
+    multiply(operand) {
+      calculator.result *= operand;
+    },
+
+    operate(operator, operand) {
+      operator(operand);
+
+      return this;
+    },
+  };
```
denys-danyliuk, line 73: save your result here
```
this.result = callback ...
```

denys-danyliuk (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+    },
+
+    subtract(operand1, operand2) {
+      return operand1 - operand2;
+    },
+
+    divide(operand1, operand2) {
+      if (operand1 === 0) {
+        return;
+      }
```
denys-danyliuk, line 60: what if your operand2 will be 0?
we can divide 0 by something, but we can't divide something by 0

denys-danyliuk (approved):

PR #1034, opened 2022-10-27
solution

IOneStarI:


src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+      return num1 * num2;
+    },
+
+    divide(num1, num2) {
+      if (num1 !== 0 && num2 !== 0) {
+        return num1 / num2;
+      } else {
+        throw new Error('Insert a non-0 value');
+      }
+    },
```
denys-danyliuk, line 73: you can't divide something by 0, that's right
but you can divide 0 by something🤓

denys-danyliuk (approved):

PR #1038, opened 2022-10-27
Solution

Anatolii-Voloshyn:


src/makeCalculator.js
```diff
@@ -53,6 +53,10 @@ function makeCalculator() {
     },
 
     divide(value, number) {
+      if (number === 0) {
+        throw new Error();
```
nazarko94, line 57: You can add in brackets a hint like: 'Not allowed to divide on zero'

nazarko94 (approved):

src/makeCalculator.js
mykola-koval: add an error description

mykola-koval (approved):
gj but add an error description

PR #1041, opened 2022-10-27
develop

Ostapchuk-Y-O:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+      return calculator.result - value;
+    },
+
+    multiply(value) {
+      return calculator.result * value;
+    },
+
+    divide(value) {
+      return calculator.result / value;
+    },
```
taraselnik, line 57: cool way! IMHO

taraselnik (approved):
enjoy

src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
  */
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      return calculator.result + value;
+    },
```
denys-danyliuk, line 45: Make it like a pure function.
For example
```
add(num1, num2) {
  return num1 + num2;
}
```

src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(method, value) {
+      this.result = method(value);
+
+      return this;
```
denys-danyliuk, line 68: save your result here
```
this.result = callback ...
```

denys-danyliuk (changes requested):

PR #1059, opened 2022-11-04
Solution

SerhiiVovchenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    },
+
+    multiply(number1, number2) {
+      return number1 * number2;
+    },
+
+    divide(number1, number2) {
+      let resultDivide = 0;
+
+      // eslint-disable-next-line no-unused-expressions
```
Bohdan-Maksymiuk, line 58: try to not disable eslint rules

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+      (number2 !== 0)
+        ? resultDivide += number1 / number2
+        : Error('Division by 0 is impossible');
```
Bohdan-Maksymiuk, line 61: if you want to do it - you should throw an error
better just create if for error case and avoid if with this ternary

Bohdan-Maksymiuk (changes requested):
Good job, take a look over comments

Bohdan-Maksymiuk (approved):

PR #1060, opened 2022-11-04
Solution

ifuride:


src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+    add(number) {
+      calculator.result += number;
```
AlisaVlasova, line 48:
```suggestion
      this.result += number;
```

AlisaVlasova (changes requested):

src/makeCalculator.js
ifuride: @AlisaVlasova if I use this suggested change, I get an error " TypeError: Cannot read property 'result' of undefined".
Please refer to the screenshot:
error

vlad-demchuk (changes requested):
if you still have questions - feel free to ask in fe_chat)

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
+      callback(number);
```
vlad-demchuk, line 43: you should use `bind` to make you `this` correct

ifuride: Done.

AlisaVlasova (approved):

PR #1064, opened 2022-11-06
Task was Solved

Stepico:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```
Svitlana-Yudina, line 57: Add check if b === 0. If it's true, an error message (can't divide by zero) should appear). Use `throw new Error`

Svitlana-Yudina (approved):
Good Job!))
Just one remark)

denys-danyliuk (approved):

PR #1068, opened 2022-11-09
Added the solution

zbyn3k:


src/makeCalculator.js
```diff
@@ -36,8 +35,41 @@
...
+
 function makeCalculator() {
-  // write code here
+  const calculatorObject = {
+    result: 0,
+
+    add(a, b) {
+      return a + b;
+    },
+
```
amitpatiljc, line 46: all methods should have only one number as a parameter , 
`add(number)`
and in this case `add` to whatever we have as a `result `is 
somthing like this 
`calculator.result += number;`

amitpatiljc (changes requested):

src/makeCalculator.js
zynkd: This has been updated. Let me know if anything else is needed. Thanks

qoa11a (approved):

PR #1080, opened 2022-11-17
solution

panianchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(a, b) {
```
Viktor-Kost, line 43: Here and for other operations as well you can use just one argument to pass to `add()` (for example "number") and then just reuse your result `this.result += number`

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
```
Viktor-Kost, line 66: And here you will pass to callback only a number

Viktor-Kost (changes requested):
Good job!

Apply some recommendations

SanyaBratashchuk (approved):

PR #1089, opened 2022-11-23
Solution

UmizDemud:


src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+      return this;
+    }
+    operate = (callback, num) => {
+      callback(num);
+
+      return this;
+    }
+  }
+
+  return new Calculator();
```
mgrinko, line 69: It works of course. But why do you need class here?
You could just return an object with methods

src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  class Calculator {
+    constructor() {
+      this.result = 0;
+    }
```
mgrinko, line 43: just
```suggestion
    result = 0;
```
Because you don't have any logic in the constructor. So you don't need it

mgrinko (approved):

src/makeCalculator.js
UmizDemud: I guess I got excited -.-

PR #1093, opened 2022-11-25
Solution

halynarydasova:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(number) {
+      calculator.result += number;
```
jstmpelowycz, line 43: generally, we should not rely on `calculator` when addressing calls
if you want to optimize it, read the article
https://www.freecodecamp.org/news/what-to-do-when-this-loses-context-f09664af076f/

jstmpelowycz (approved):

src/makeCalculator.js
halynarydasova: Thanks

PR #1095, opened 2022-11-25
solution

ysrckr:


lefteris-L (approved):
I don't think it gets better than this 👍

naziamoff (approved):
🔥 🚀 
Would look more neat if methods are separated with an empty line

PR #1096, opened 2022-11-26
Solution

vitalii-borzov:


src/makeCalculator.js
```diff
@@ -38,6 +38,38 @@
...
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+
+    add(a) {
+      calculator.result += a;
```
AlisaVlasova, line 50:
```suggestion
      this.result += a;
```

AlisaVlasova (changes requested):

src/makeCalculator.js
vitalii-borzov: If I change calculator.result += a; for this.result += a; it's does not working

vitalii-borzov: Снимок экрана 2022-11-26 165634

nazarmatsevych (approved):
👍

PR #1098, opened 2022-11-26
calculator Valentos

mr-valentos:


llllewvvaa (approved):
Everything is good, but, I think, it's better to add spaces between methods

valerii-derkach (approved):

PR #1099, opened 2022-11-26
Solution

nurdanokcu:


CharAnagn (approved):
Perfect 🔥

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(a, b) {
+      return a + b;
+    },
```
naziamoff, line 45: why don't you just operate with the result variable right in this method?

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
```
naziamoff, line 66: this way you won't need to pass `this.result` each time

naziamoff (approved):
Well done, but consider my comments

src/makeCalculator.js
nurdanokcu: Okay I will implement the suggestions. Thank you.

PR #1103, opened 2022-11-28
Done

SalehAkaJim:


IoghiStudio (approved):
Awesome!

naziamoff (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(val) {
+      this.result += val;
+
+      return this;
```
naziamoff, line 58: you return this inside operate method, so this makes the chain call possible itself

naziamoff (approved):
🔥

src/makeCalculator.js
SalehBUD: Fixed 😅

PR #1113, opened 2022-11-29
solution v1

Lun4trick:


IoghiStudio (approved):
Look good!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
+    },
```
naziamoff, line 69: the operation methods have access to this.result themselves, no need to pass this.result to them explicitly

naziamoff (approved):

PR #1115, opened 2022-11-29
Solution

OlggaJul:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculations = {
+    result: 0,
+    add(n) {
+      return calculations.result + n;
+    },
```
superpooperxxx, line 44: Better not use calculation.result, because the variable name can change and ruin everything. Consider accepting result as a parameter.

superpooperxxx (changes requested):

superpooperxxx (approved):
All good🔥

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      return previousResult * n;
+    },
+    divide(previousResult, n) {
+      return previousResult / n;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
```
Ihor-Karpyn, line 58: Let's add an empty lines between methods to increase readable

Ihor-Karpyn (approved):

PR #1118, opened 2022-11-30
Solution

llllewvvaa:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
```
mr-valentos, line 41: 2 [NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)

mr-valentos (approved):

valerii-derkach (changes requested):
fix comments, please

src/makeCalculator.js
llllewvvaa: Thanks for the hint, but the `result` are written in the tests, so it can't be another var

valerii-derkach (approved):

PR #1123, opened 2022-12-01
Solution

Yuriy-Khoptyanii:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const Calculator = {
+    result: 0,
+
+    add(a, b) {
```
anastasiia-bilska, line 43: it is better to give a more understandable name for the parameters

anastasiia-bilska (changes requested):
Good, but please read my comment 👍

src/makeCalculator.js
Yuriy-Khoptyanii: I think that is enough for this case and not important because we can read this too easy.
But we always can improve our code - that's why I changed it, thanks.

anastasiia-bilska (approved):
great!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const Calculator = {
```
valerii-derkach, line 40:
```suggestion
  const сalculator = {
```

valerii-derkach (changes requested):

solaryasha (approved):
Good job!

PR #1128, opened 2022-12-01
solution

Alireza-Mohammadhossein:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, num) {
+      callback.call(this, num);
+
+      return this;
```
naziamoff, line 68: do you need to pass `this` explicitly here? To me, it seems that all the operations already have access to `this`, due to they're in the same object

naziamoff (approved):

PR #1131, opened 2022-12-02
first attempt at Advanced Calculator, with callback

Kavster92:


src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const calc = {
```
naziamoff, line 41: n n n pls dnt mak unncsry abbrs in ur vars, it mks ur cod mor cmplcted t read

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
```
naziamoff, line 67: do you need to pass `this` explicitly here? To me it seems that your methods already have an access to `this`

naziamoff (approved):
LGTM, just a couple of suggestions

PR #1133, opened 2022-12-02
Solution

IrynaK-a:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(number) {
+      this.result += number;
+
+      return this.result;
```
ivanchuk-yurii, line 46:
```suggestion
      return this.result + number;
```
anyway you rewrite `this.result` in `operate`. fix for all methods

ivanchuk-yurii (changes requested):

qoa11a (approved):

PR #1145, opened 2022-12-05
solution

kenvastolord:


src/makeCalculator.js
```diff
@@ -38,6 +38,33 @@
...
+    },
+    divide(number) {
+      calculator.result /= number;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+    operate(fn, n) {
```
Iksich, line 60: You could use callback instead of fn cause it's a bit unclear.

meforoma (approved):

PR #1147, opened 2022-12-05
solution

vladyslav-krainyk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(firstNumber, secondNumber) {
```
SanyaBratashchuk, line 43:
```suggestion
    add(number) {
```
Try to make it work with one param

SanyaBratashchuk (changes requested):

stanislav-topikha (approved):

PR #1155, opened 2022-12-13
commit

OleksChernikov:


src/makeCalculator.js
```diff
@@ -36,8 +36,32 @@
  *
  * @return {object}
  */
-function makeCalculator() {
-  // write code here
-}
+  function makeCalculator() {
+    let obj = {
+      result: 0,
```
polosanya, line 41: image
```suggestion
    const obj = {
      result: 0,
```
Let's use more descriptive names, not just `obj` and `result`.
Also, you don't change the object, so use const instead of let.
Or you can even just immediately return the object

src/makeCalculator.js
```diff
@@ -36,8 +36,32 @@
...
+      add: function(number) {
+        obj.result = (obj.result + number);
+      },
+      subtract: function(number) {
+        obj.result = (obj.result - number);
+      },
```
polosanya, line 47:
```suggestion
      subtract(number) {
        obj.result -= number;
      },
```
image
image

polosanya (changes requested):
Almost done! Don't forget to check yourself with the checklist before submitting)

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+    add(number) {
+      obj.result += number;
+    },
+    subtract(number) {
```
Anastasiia-Svintsova, line 45: Better to add spaces to improve the readability of the code
```suggestion
    add(number) {
      obj.result += number;
    },
    
    subtract(number) {
```

Anastasiia-Svintsova (approved):

PR #1158, opened 2022-12-13
initial commit

ihorivna007:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    operate: function(callback, value) {
+      this.result = callback(value, this.result);
+
+      return this;
+    },
+
+    add: function(value, result) {
+      return result + value;
+    },
```
SanyaBratashchuk, line 51: image
Fix it everywhere

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    subtract: function(value, result) {
+      return result - value;
+    },
```
SanyaBratashchuk, line 55:
```suggestion
    subtract: function(value) {
      return this.result - value;
    },
```
Try to make it work like that

SanyaBratashchuk (changes requested):

TarasHoliuk (approved):

PR #1160, opened 2022-12-14
solution

Nikoramo:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    add(a, b) {
+      return (a + b);
```
UnknownAI, line 42: everything is good, just small advice - it can be simplified by omitting the parentheses

UnknownAI (approved):

PR #1163, opened 2022-12-16
solution

Roman-Malysh:


src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
  */
 function makeCalculator() {
   // write code here
+
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      return calculator.result + number;
```
AlisaVlasova, line 46:
```suggestion
      return this.result + number;
```

AlisaVlasova (changes requested):

src/makeCalculator.js
Roman-Malysh: Do not pass tests. I tried everything, just approve please

AlisaVlasova: try 
this.result += number;
return this;

AlisaVlasova (changes requested):

PR #1171, opened 2022-12-23
Solution

RomaSheva1987:


src/makeCalculator.js
```diff
@@ -38,6 +38,27 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => a / b,
```
AlisaVlasova, line 46: these methods should take one argument and add/divide/etc. to the result and return this

AlisaVlasova (changes requested):

tetiana-ivanchenko (approved):

PR #1177, opened 2022-12-26
Calculator is done

kateryna-stemblivska:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
+    result: 0,
+
+    add() {},
+    subtract() {},
+    multiply() {},
+    divide() {},
```
loralevitska, line 46: perform operations in the corresponding functions, remove `switch case`

loralevitska (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
+    result: 0,
```
polosanya, line 41:
```suggestion
  const calculator = {
    result: 0,
```
image

polosanya, line 41:
```suggestion
  const calculator = {
    result: 0,
```
Let's use more meaningful names)
image

polosanya (changes requested):
Almost perfect!

loralevitska (approved):

PR #1180, opened 2022-12-29
add solution

Anastasiia-Khudych:


Kardash-Oleksandr (approved):
Good work)

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(action, number) {
+      action(number);
+
+      return calculator;
```
ArturErmolenko, line 46: Each `operate` call takes a callback and a number and sets the appropriate value to the result property.
you should set needed value to the `result` in operate method

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1181, opened 2022-12-30
Create calculator

OlhaYevstifieieva:


Malva37 (approved):
looks great)

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
```
ArturErmolenko, line 68: Each `operate` call takes a callback and a number and sets the appropriate value to the result property.
you should set needed value to the `result` in operate method

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1183, opened 2022-12-31
Solution

Kardash-Oleksandr:


sv-dubov (approved):
Good job! But I think its better to add spaces between functions. It will be more readable.

Kardash-Oleksandr:
> Good job! But I think its better to add spaces between functions. It will be more readable.

Thanks. Fixed)

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    multiply(number) {
+      calculator.result *= number;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
```
ArturErmolenko, line 68: Each `operate` call takes a callback and a number and sets the appropriate value to the `result `property.

ArturErmolenko (changes requested):

src/makeCalculator.js
Kardash-Oleksandr: fixed

PR #1184, opened 2022-12-31
Solution

Avichez:


ihor-soloviov:
Good job!

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(number) {
+      this.result += number;
+    },
```
Anna-Tkachenko, line 45: check task description: each `operate` call takes a callback and a number and sets the appropriate value to the `result` property.
separate calculate functions shouldn't know anything about the object or changing some of calculator properties
they just should accept needed numbers and provide needed result

Anna-Tkachenko (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(prevNumber, number) {
```
Anna-Tkachenko, line 43: will be better to create more general variable names
it's not clear why one of numbers is previous, and other one not

Anna-Tkachenko (approved):

PR #1185, opened 2022-12-31
Calculator solution

andrieieva-yuliia:


NVyrsta (approved):
Good job :)

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
```
Anna-Tkachenko, line 69: `operate` function should take needed arguments, provide corresponding action and set the result value to the result property

other methods shouldn't know anything about some `result` property, they should just accept 2 numbers and provide correct result

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #1189, opened 2023-01-02
Make Calculator - completed

GarryRocksman:
Have a nice day!!!

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num, context) {
+      context.result += num;
+    },
+
+    subtract(num, context) {
+      context.result -= num;
+    },
+
+    multiply(num, context) {
+      context.result *= num;
+    },
+
+    divide(num, context) {
+      context.result /= num;
+    },
+
+    operate(callback, num) {
+      callback(num, this);
+
+      return this;
+    },
+
+    reset() {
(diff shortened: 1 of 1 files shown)


tania-boiar (approved):
Interesting solution!
Well done! 🔥

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+    divide(num, context) {
+      context.result /= num;
+    },
+
+    operate(callback, num) {
+      callback(num, this);
+
+      return this;
+    },
```
Anna-Tkachenko, line 63: check task description: Each `operate` call takes a callback and a number and sets the appropriate value to the `result` property.

Anna-Tkachenko (changes requested):

src/makeCalculator.js
GarryRocksman: fixed

GarryRocksman: thank you for reviewing, I will try be more concentrated when reading the task

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num, res) {
```
Anna-Tkachenko, line 43: it's not clear why one of the operand is `num`, and second one is `res`, will be better to name them in more general way without any relation to some number or result property

Anna-Tkachenko (approved):
Good job! 🔥

PR #1191, opened 2023-01-03
makeCalculator function was created

svitjojo:


NVyrsta (approved):
Good job :)

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number) {
+      callback(number);
+
+      return this;
```
ArturErmolenko, line 45: check task description: Each `operate` call takes a callback and a number and sets the appropriate value to the `result` property.

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1192, opened 2023-01-03
Solution

tania-boiar:


Kardash-Oleksandr (approved):
You can make the code easier to read by passing a single argument to the method and working directly with calculate.result inside the method.

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num, currentNum) {
```
Anna-Tkachenko, line 43: will be better to use more general names for arguments, it's not clear why one of them is `num`, and the second is `currentNum`

Anna-Tkachenko (approved):
Good job!

PR #1193, opened 2023-01-03
Solution

chermak-oleh:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    divide(number) {
+      calculator.result /= number;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
```
tania-boiar, line 62: You could add `calculator.result` as a second parameter to your callback function in `operate method`,
so you don't have to change `calculator.result` in every operating method
Like this:
```
    divide(result, number) {
      return result / number;
    },

  operate(callback, number) {
```
        
I'm not sure about naming of variables but hope that you could get the idea.

tania-boiar (approved):
Good job! 🔥
Left you a comment. Look through it, if you are interested.

src/makeCalculator.js
Anna-Tkachenko: agree, fix @tania-boiar comment
methods for working with numbers shouldn't know anything about some object or some property
they just should accept 2 numbers and provide needed result

Anna-Tkachenko (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(result, number) {
```
Anna-Tkachenko, line 43: will be better to name variables in more general name
it's not clear why one of arguments is the result and the other is not

Anna-Tkachenko (approved):
Good job!

PR #1200, opened 2023-01-04
Solution

Kravchenko71213:


Kardash-Oleksandr (approved):
You can make the code easier to read by passing a single argument to the method and working directly with calculate.result inside the method.

ArturErmolenko (approved):

PR #1201, opened 2023-01-04
solution

Oleksandr-Maly:


Kardash-Oleksandr (approved):
An unusual solution, but you can simplify the readability of the code if you pass a single argument to the method and work directly with calculate.result inside the method.

ArturErmolenko (approved):

PR #1203, opened 2023-01-04
Solution

maasiimka:


ihor-soloviov:
Very nice solution. I spent a lot of time to understand why 'this' not works right, so made my solution in another way. But now I see how I need to work. Good job!

PR #1205, opened 2023-01-04
Advanced calculator solution

PolinaSukhorukova:


Kardash-Oleksandr (approved):
Nice work)

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return calculator;
```
ArturErmolenko, line 68: Each `operate` call takes a callback and a number and sets the appropriate value to the` result `property.

ArturErmolenko (changes requested):

PR #1206, opened 2023-01-04
Task was solved

ihor-soloviov:


Kardash-Oleksandr (changes requested):
You don't have to return a result after every operation. And I'm not sure that there is a need to set default values ​​to methods.

ArturErmolenko (approved):

PR #1207, opened 2023-01-04
Solution

vanst1:


Kardash-Oleksandr (changes requested):
You dont need to return calculator after every operation

src/makeCalculator.js
```diff
@@ -36,8 +36,39 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operator, number) {
+      operator(number);
+
+      return this;
```
ArturErmolenko, line 67: check task description: Each `operate` call takes a callback and a number and sets the appropriate value to the `result` property.

ArturErmolenko (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #1208, opened 2023-01-04
Calculator

OlehSharha:


src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    },
+
+    operate(callback, x) {
+      callback(x);
+
+      return calculator;
+    },
+  };
+
+  return calculator;
```
GarryRocksman, line 67: check task description: Each operate call takes a callback and a number and sets the appropriate value to the result property.
you should set needed value to the result in operate method but you  setting it in each callback that's not correct, try ti fix it

GarryRocksman (changes requested):
You did great work but your code need some fixes,  try to read requirement one mote time

PR #1211, opened 2023-01-05
solution_1

Borys-Andrew:


0xNazarii (approved):

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      this.result /= value;
+    },
+
+    operate(callback, value) {
+      callback.bind(this)(value);
```
Denchepornyuk, line 67: It's better to pass `this.result` inside callback function as an argument instead of using bind

denys-cheporniuk (changes requested):

src/makeCalculator.js
Borys-Andrew: fixed

0xNazarii (commented):
❤️

alexandra-protyanova (approved):
Well done 👍

PR #1213, opened 2023-01-09
Finish task

U-Ra-All:


src/makeCalculator.js
```diff
@@ -38,6 +38,47 @@
...
   // write code here
+  const calculator = {};
+
+  calculator.result = 0;
+
+  calculator.add = (num) => {
+    calculator.result += num;
+
+    return calculator.result;
+  };
```
IvanFesenko, line 49: You should add methods and static fields while initialising object
```suggestion
  const calculator = {
    result = 0;
    add (num) {
      //...
    }
    //... 
  };

```

IvanFesenko (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,34 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    add(num) {
+      this.result += num;
+      return this.result;
+    },
+    subtract(num) {
```
OlEzhA0, line 47: add spaces between blocks
```suggestion
    },

    subtract(num) {
```

OlEzhA0 (approved):
good job

PR #1218, opened 2023-01-11
Solution

novak-k:


src/makeCalculator.js
```diff
@@ -36,8 +36,48 @@
...
+
+  };
+
+  function makeOperate(operation, value) {
+    ops.result = operation(ops.result, value);
+
+    return ops;
+  }
+
+  function makeAdd(a, b) {
```
TarasHoliuk, line 58: write it (all functions) as methods and use `this.result` instead of passing the second argument

TarasHoliuk (changes requested):

src/makeCalculator.js
novak-k: Please give me a hint, I didn't get it, what I suppose to assign to this? 
When I set ops to this it throws me an error

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
  */
+
+const ops = {
+  result: 0,
+  // 'add': makeAdd,
+  // 'subtract': makeSubst,
+  // 'multiply': makeMult,
+  // 'divide': makeDiv,
+  // 'reset': makeReset,
+  // 'operate': makeOperate,
```
vlad-demchuk, line 47: remove comments

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+
+};
+
+ops.operate = function makeOperate(operation, value) {
```
vlad-demchuk, line 51: better to move methods to the object and use descriptive variable name for it, smth like `calc`

vlad-demchuk (changes requested):

src/makeCalculator.js
vlad-demchuk: if you have any questions, just feel free to ask in `fe_chat`

src/makeCalculator.js
novak-k: done

PR #1219, opened 2023-01-11
solution

drhtka:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    multiply(number) {
+      return this.result * number;
+    },
+
+    divide(number) {
+      return this.result / number;
+    },
+
+    operate(operation, number) {
+      this.result = operation.call(this, number);
```
loralevitska, line 60:
```suggestion
      this.result = operation.call(this.result, number);
```

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      return this.result + number;
+    },
```
loralevitska, line 45: It will return to `this.result` in 60 line. Just do operation between to parameters. 
image
```suggestion
    add(a, b) {
      return a + b;
    },
```

Fix all cases

SanyaBratashchuk (approved):

PR #1224, opened 2023-01-12
added solution

constantweb:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(a, b) {
+      return a + b;
```
tetiana-ivanchenko, line 56: Each `operate` should takes a callback and a number and sets the appropriate value to the `result` property.

tetiana-ivanchenko (changes requested):

vlad-demchuk (approved):

PR #1229, opened 2023-01-18
Calculator

Igor-Yaremchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
```
polosanya, line 41: image

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    add(result, value) {
+      return result + value;
+    },
```
polosanya, line 45: You already have a result in the upper scope
```suggestion
    add(a, b) {
      return a + b;
    },
```

polosanya (changes requested):
Let's polish it up)

src/makeCalculator.js
polosanya: Not fixed

polosanya (approved):
I'm approving it, but please pay attention to the last comment

PR #1231, opened 2023-01-19
Solution

pushkarskiyrodion:


src/makeCalculator.js
```diff
@@ -36,8 +36,47 @@
...
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, value) {
+      return callback.call(this, value);
```
loralevitska, line 45: You may just call operation and then return `this` here. It helps you avoid code duplicate in each operation (you returned `this`)

src/makeCalculator.js
```diff
@@ -36,8 +36,47 @@
...
+      this.result = 0;
+
+      return this;
+    },
+
+    add(value) {
+      this.result += value;
+
+      return this;
+    },
```
loralevitska, line 58: If you do as in the first comment, this code can be shortened and your code style will be cleaner. Fix this in all methods

```suggestion
    add(value) {
      this.result += value;
    },
```

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

DarkMistyRoom (approved):

PR #1235, opened 2023-01-23
Solution makeCalculator

NazariiAlieksieiev:


src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
  */
 function makeCalculator() {
   // write code here
```
VitaliyBondarenko1982, line 40: comments could be removed

VitaliyBondarenko1982 (approved):

PR #1243, opened 2023-01-27
Task done

andriy-litkovets:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const answer = {
+    result: 0,
+    operate: function(callback, number) {
```
agniya-i, line 42:
```suggestion
    operate(callback, number) {
```
not a big mistake, but you can define function in a more simple way

agniya-i (approved):

PR #1244, opened 2023-01-27
Solution

Serhii-Shtokalo:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
```
ViktorHavryliuk, line 40:
```suggestion
  const calculator = {
```

ViktorHavryliuk (approved):
Good work, but it's better to give variables descriptive names

PR #1245, opened 2023-01-27
task was solved

vanyachyzh:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+      this.result = callback(this.result, number);
+
+      return this;
+    },
+    reset: function() {
+      this.result = 0;
+
+      return this;
+    },
+    add: function(result, number) {
```
AlisaVlasova, line 52:
```suggestion
    add: function(number) {
```
these methods should not receive result as a prop, use this.result inside

AlisaVlasova (changes requested):

IvanFesenko (approved):

PR #1252, opened 2023-01-30
solution

44shaggy:
I wonder why "this" keyword didnt work in other methods.


qoa11a (approved):
you should ask questions in chat

PR #1259, opened 2023-02-01
Calculator 2.0

oleksandr-plias:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
+    },
```
loralevitska, line 45: Good job! Try to rewrite your solution with `this`
```suggestion
    add(number) {
      this.result += number;
    },
```

loralevitska (changes requested):

tetiana-ivanchenko (approved):

PR #1260, opened 2023-02-02
Solution

EvgenyNikitinEvgeny:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
+    },
```
loralevitska, line 45: Try to use `this` 
```suggestion
    add(number) {
      this.result += number;
    },
```

loralevitska (changes requested):
Good job!
Let's rewrite your solution for more practice with `this` in JS

tetiana-ivanchenko (approved):

PR #1262, opened 2023-02-02
solution calc advanced

Spec0s:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    reset() {
+      calculator.result = 0;
```
polosanya, line 43:
```suggestion
      this.result = 0;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+
+      return this;
+    },
+    add(a) {
+      calculator.result += a;
```
polosanya, line 48:
```suggestion
      this.result += a;
```
Fix all cases

polosanya (changes requested):

src/makeCalculator.js
Spec0s: if i put this instead of calculator i get errors  . before was doing with this and didnt figure out why its not working so had to put calculator

Spec0s: i had to find how call() works and call() js function in general, or i missed video lecture where it was mentioned ?

polosanya: You could also pass this.result as an argument of operate and add a value to it. 
Or save the result of operate function into this.result. 
I guess you will have theory and videos about call, bind and apply in future sections, but here is some additional theory for you, if you want to know more :)
https://www.freecodecamp.org/news/understand-call-apply-and-bind-in-javascript-with-examples/

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      this.result -= a;
+    },
+    multiply(a) {
+      this.result *= a;
+    },
+    divide(a) {
+      this.result /= a;
+    },
+    operate(operation, number) {
+      operation.call(this, number);
```
polosanya, line 60: 🔥

polosanya (approved):
Well done!

PR #1269, opened 2023-02-07
Calculator solution

TkachYuliia:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(x) {
+      return calculator.result + x;
+    },
```
IvanFesenko, line 45: Let's rewrite your solution for more practice with `this` in JS.
Add `x` is not clear name. Use `value` or `number` insteat
```suggestion
      return this.result + x;
    },
```

IvanFesenko (changes requested):
Good job. Lets do some refactor

OlEzhA0 (approved):

PR #1270, opened 2023-02-08
solution

ivanm2706:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
 function makeCalculator() {
-  // write code here
+  const resultObj = {
+    result: 0,
+    add(a, b) {
+      return b + a;
+    },
+    subtract(a, b) {
+      return b - a;
+    },
```
VitaliyBondarenko1982, line 47:
```suggestion
    add(a, b) {
      return b + a;
    },
    
    subtract(a, b) {
      return b - a;
    },
```

VitaliyBondarenko1982: Better to separate methods with empty line for better reading code

VitaliyBondarenko1982 (approved):

PR #1288, opened 2023-02-16
Solution

Vasyl-Pauk:


src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
```
```suggestion
  return {
```
you may return it immediately

TarasHoliuk: FYI

TarasHoliuk (approved):
:+1:

PR #1291, opened 2023-02-17
commit message

Peti51:


src/makeCalculator.js
```diff
@@ -36,8 +36,51 @@
...
+      calculator.result *= n;
+    },
+
+    reset: function() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    result: 0,
```
Volodymyr-Melnyk-Fron10der, line 68: I suggest to move reset property to the top

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, looks good, except for a few moments here:

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+      calculator.result = 0;
+
+      return this;
+    },
+
+    add: function(n) {
+      calculator.result += n;
+    },
+
+    subtract: function(n) {
```
polosanya, line 52:
```suggestion
    add(n) {
      calculator.result += n;
    },

    subtract(n) {
```
Fix all cases
image

polosanya (changes requested):

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
  */
+
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
```
Volodymyr-Melnyk-Fron10der, line 45: You can access object with `this`
```suggestion
      this.result = 0;
```

Volodymyr-Melnyk-Fron10der (approved):
👍

PR #1293, opened 2023-02-17
Solution

iryna-serniuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      calculator.result -= number;
+    },
+
+    multiply(number) {
+      calculator.result *= number;
+    },
+
+    divide(number) {
+      calculator.result /= number;
+    },
```
IvanFesenko, line 63: You should use this.result here
```suggestion
    add(number) {
      this.result += number;
    },

    subtract(number) {
      this.result -= number;
    },

    multiply(number) {
      this.result *= number;
    },

    divide(number) {
      this.result /= number;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      callback(number);
+
```
IvanFesenko, line 45: To avoid error use `callback.call()`

IvanFesenko (changes requested):
Good work. Check my comments

VitaliyBondarenko1982 (approved):

PR #1296, opened 2023-02-17
calculator was made

lilia-obushenko:


vladyslav-mosor (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+  const reset = function() {
+    this.result = 0;
+
+    return this;
+  };
+
+  const calculator = {
+    result: 0,
+    reset,
+    operate,
```
OleksiiBlezniuk, line 55: Ok, but why to move it to constants? It's enough to just add methods to objects

OleksiiBlezniuk (approved):

src/makeCalculator.js
lilia-obushenko: Fixed)

PR #1299, opened 2023-02-18
solution

VadimDrobyazko:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```
IvanFesenko, line 57: Try to refactor:
You methods should get 1 arg number and use this

```
add(number) {
  return this.result + number;
}, 
```

IvanFesenko (approved):

PR #1304, opened 2023-02-19
Advanced Calculator - Solution

Yevhenii0536:


sergey-lipikhin (approved):
GJ

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, item) {
+      this.result = callback(this.result, item);
```
OleksiiBlezniuk, line 50: Ok, but what is `item`? I guess `number` will be more understandable in this case (it's a number that we want to use for calculation)

OleksiiBlezniuk (approved):

PR #1306, opened 2023-02-20
Calculator

Yevhen-Ivanchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+
+    add: (a, b) => {
+      return a + b;
+    },
```
Anastasiia-Svintsova, line 45: Try to refactor:
You methods should get 1 arg number and use this. And you don't really need to return the value

```
add(number) {
  this.result += number;
},

File: src/makeCalculator.js [hunk 2]
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
```
Anastasiia-Svintsova, line 40: Give a more meaningful name, please 
```suggestion
  const calculator = {
```

Anastasiia-Svintsova (changes requested):

src/makeCalculator.js
```diff
@@ -1,43 +1,43 @@
...
+      this.result = callback(number);
+
+      return this;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
```
witflash, line 17: If you separate methods with an empty line use this approach everywhere.

witflash (approved):
Well done 👍

qoa11a (approved):

PR #1309, opened 2023-02-20
calculator was created

Yekatery:


gretoq (approved):
GJ! But 'num' is not good name for variables, 'num1' and 'num2' also.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num1, num2) {
```
OleksiiBlezniuk, line 43:
```suggestion
    add(a, b) {
```

OleksiiBlezniuk: Don't use numbers in variables names
In this case if all that you know about this values - is that they're numbers (without any context of it's usage) and they'll be used for mathematical operations - it's better to use common math names, like `a` or `b`

OleksiiBlezniuk: Or even better, we do know that first param - is a current result of operation, and second - is a number to add/divide/multiply/subtract, so you can use `result` and `number` here
But never use numbers in variable names

OleksiiBlezniuk (changes requested):
GJ, but there is a few issues in variables naming to fix

sergii-nosachenko (approved):
Good work 👍

PR #1311, opened 2023-02-20
calculator task is done

stanislavkulakovskyi:


BozhenaBodnar (approved):
good work

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    multiply(result, num) {
+      return result * num;
+    },
+
+    reset() {
+      this.result = 0;
+
+      return calculator;
```
OleksiiBlezniuk, line 62: ok, but what if we'll rename this object? Let's return `this` instead

OleksiiBlezniuk (approved):

PR #1324, opened 2023-02-22
Solution for Make Calculator

ctk12:


src/makeCalculator.js
```diff
@@ -38,6 +38,33 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    add(value) {
+      calculator.result += value;
+    },
+    subtract(value) {
```
polosanya, line 46:
```suggestion
    result: 0,
    
    add(value) {
      calculator.result += value;
    },
    
    subtract(value) {
```
Don't forget to provide empty lines between methods, so it is easier to read your code

polosanya (approved):
Good job!

PR #1328, opened 2023-02-22
calculator

VadKol:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+  return {
+    add,
+    subtract,
+    multiply,
+    divide,
+    reset,
+    operate,
+    get result() {
+      return result;
+    }
```
sergii-nosachenko, line 77: A smart combination of getter and closure, but the task requires an object to have property, not a getter.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  function add(num) {
```
sergii-nosachenko, line 42:
```suggestion
  function add(value) {
```
Use meaningful argument name

sergii-nosachenko (approved):
Good work 👍
Commented too.

PR #1329, opened 2023-02-23
Solution

vitalii-heiko:


src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
  */
 function makeCalculator() {
   // write code here
+
+  const calculator = {
```
Anastasiia-Svintsova, line 42: Don't leave unnecessary comments
```suggestion
function makeCalculator() {
  const calculator = {
```

src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
...
+    result: 0,
+    operate(callback, number) {
+      this.result = callback(this.result, number);
+
+      return this;
+    },
+
+    add(result, number) {
+      return result + number;
```
Anastasiia-Svintsova, line 51: You should pass only the number here. And seems that return is redundant here
```suggestion
    add(number) {
      this.result += number;
```

Anastasiia-Svintsova (changes requested):

agniya-i (changes requested):
There are no changes. 
Please fix comments and push changes

polosanya (approved):

PR #1342, opened 2023-02-28
solution :)

katerynashylina:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(operation, number) {
+      operation.call(this, number);
+
+      return this;
+    },
```
SerhiiKirik, line 47: Nice solution
also, you can do like that

```suggestion
    operate(callback, number) {
      this.result = callback(number)

      return this;
    },
```

SerhiiKirik (approved):
Good job 🔥

PR #1344, opened 2023-03-01
advanced_calculator

OlegRak86:


src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    'result': 0,
+    multiply(x) {
+      obj.result *= x;
+
+      return obj.result;
```
vlad-demchuk, line 45: let's use `this`
return this.result *= x

vlad-demchuk (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    'result': 0,
```
slavanosatskyi, line 41: Minor mistake
```suggestion
    result: 0,
```

slavanosatskyi (approved):

PR #1352, opened 2023-03-07
complete task

VladGraund:


src/makeCalculator.js
```diff
@@ -36,8 +36,33 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, value) {
```
OlEzhA0, line 43: would be better to add empty line between blocks

OlEzhA0 (approved):

PR #1357, opened 2023-03-10
Solution

Dmytro-Kretov:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(result, a) {
+      return result + a;
+    },
+    subtract(result, a) {
+      return result - a;
+    },
```
```suggestion
    add(result, a) {
      return result + a;
    },
    
    subtract(result, a) {
      return result - a;
    },
```

sTorba24: use blank lines between methods, it makes code more readable

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(result, a) {
```
sTorba24, line 42:
```suggestion
    add(a, b) {
```

sTorba24: using the same name as field above is not the best idea

sTorba24 (approved):
Good work! A few comments to avoid in future :)

PR #1369, opened 2023-03-18
create function  makeCalculator

VenomMB:


igorkruz (approved):
Good job 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(x) {
+      calculator.result += x;
```
igorkruz, line 43: x is not a good name for the variable, it's better to change it to value or another appropriate name

PR #1374, opened 2023-03-19
Advanced calculator online

roshirin:


roshirin:
> Everything is great! Perhaps only change the name of the 'result' variable (according to checklist #2 ([NAMING] - use proper names for variables - if you want to name variable result, pick a better one :)))

I totaly agree but in this case the 'result' property is described in the task as a valid object field. And I decided to leave it as it is.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      return this;
+    },
+
+    operate(callback, b) {
+      callback.call(this, b);
+
+      return this;
+    },
+
+    result: 0,
```
Olexandr6, line 69: Pay attention to order of properties in objects, they should be first and then methods

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, b) {
+      callback.call(this, b);
```
Olexandr6, line 64: 🔥

Olexandr6 (approved):

PR #1380, opened 2023-03-20
calc

Stambuliak:


prokivskaaa (approved):
LGTM🦾

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      callback.call(this, value);
```
liliia-livandovskaya, line 64: `value` is a bad name for the variable

liliia-livandovskaya (approved):

PR #1384, opened 2023-03-20
Solution

92BS:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    divide(b) {
+      this.result = this.result / b;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(method, b) {
+      method.call(this, b);
```
polosanya, line 60: Why exactly `b`?)
Can't it be just `number`?

polosanya (approved):
Good job!

PR #1385, opened 2023-03-21
solution

glots-a:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(x) {
+      calculator.result += x;
```
Hamuud, line 44: You can use this instead of calling the calculator variable every time

Hamuud (approved):
Well done, but If I were you, I change 'calculator.result' to 'this.result'

Olexandr6 (approved):

PR #1386, opened 2023-03-21
Create calculator with callback

Hamuud:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+      this.result += value;
+    },
+
+    subtract(value) {
+      this.result -= value;
+    },
+
+    multiply(value) {
+      this.result = this.result * value;
+    },
```
igorkruz, line 51: you can write it using a short form: this.result *= value

src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+
+    divide(value) {
+      this.result = this.result / value;
+    },
```
igorkruz, line 55: the same here

igorkruz (approved):
good job 👍
I think with the suggested change your code would be better

src/makeCalculator.js
liliia-livandovskaya: agree

liliia-livandovskaya (changes requested):
Let's rewrite your code according to the checklist p.3

liliia-livandovskaya (approved):
🔥

PR #1387, opened 2023-03-21
done, but Misha unsaid about call/apply...(((

victor-buldenko:


SerhiiKirik (approved):
Well done 🚀 
You can avoid call / apply, 
If you just create obj & write obj.result = callback(num)

victor-buldenko:
> obj.result = callback(num)

how its posible? If all methods includes 'this.properties... do something' you can call `callback(num)` without target=this;
show me plz, what are you mean, i don`t get it (((

PR #1391, opened 2023-03-21
Solution

Anna-Kuzmenko:


Anna-Kuzmenko:
Why doesn't this function work with this.result, only calculator.result?

liliia-livandovskaya (approved):
Awesome 👍

PR #1392, opened 2023-03-21
create solution with makeCalculator()

kukla1989:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    divide(a, b) {
+      return a / b;
+    },
+
+    operate(callback, num) {
+      this.result = callback(this.result, num);
+
+      return this;
+    },
```
liliia-livandovskaya, line 63:
```suggestion
    operate(callback, num) {
      operate(callback, num) {
      callback.call(this, num);

      return this;
    },
```
What about your question on Q&A, you can use `call` to set the context manually. Then you would be able to use `this` in methods ` add, subtract ...`

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+  return {
+    result: 0,
+
+    add(a, b) {
+      return a + b;
+    },
+
+    subtract(a, b) {
+      return a - b;
+    },
```
liliia-livandovskaya, line 49:
```suggestion
   subtract: (a, b) => a - b;
```
to make it look simpler

liliia-livandovskaya (approved):

PR #1393, opened 2023-03-21
Solution

Goto-Jesus:


Kavon-Skela (approved):
Good job overall! Maybe, it will be better to choose a better variable name than "value"

PR #1395, opened 2023-03-21
Solution

Kavon-Skela:


kukla1989 (approved):
good job

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      return this;
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
```
liliia-livandovskaya, line 73: `value` is a bad name for the variable

liliia-livandovskaya (approved):

PR #1400, opened 2023-03-24
https://github.com/Vitalii-Maistrenko/js_task-guideline/blob/master/README.md

Vitalii-Maistrenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,64 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
```
polosanya, line 40: image

src/makeCalculator.js
```diff
@@ -37,7 +37,64 @@
...
+    add,
+    subtract,
+    multiply,
+    divide,
+    reset,
+    operate,
+    get result() {
+      return result;
+    },
+  };
```
polosanya, line 80: It is better to initially create an object with all these methods and return it here

src/makeCalculator.js
```diff
@@ -37,7 +37,64 @@
...
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  function add(num) {
+    result += num;
+  }
+
+  function subtract(num) {
```
polosanya, line 46: Make all functions be methods of object calculator
image

src/makeCalculator.js
```diff
@@ -37,7 +37,64 @@
...
+// console.log(calculator.result); // 0
+
+calculator
+  .operate(calculator.add, 10)
+  .reset()
+  .operate(calculator.subtract, 20)
+  .operate(calculator.divide, 5)
+  .operate(calculator.multiply, 7);
+
+// console.log(calculator.result); // -28
```
polosanya, line 98: You can remove this

polosanya (changes requested):

PR #1401, opened 2023-03-24
first version

Vladimir-Zadorozhnyi:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
+    result: 0,
+    add(value) {
+      calculator.result += value;
```
agniya-i, line 43: almost correct, you should use `this` 

```suggestion
      this.result += value;
```

agniya-i (changes requested):

etojeDenys (approved):
looks good

PR #1402, opened 2023-03-26
solution

iHateAntichrist:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
```
jstmpelowycz, line 40: `result` should be a part of object you return

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+    result = 0;
+
+    return calculator;
+  }
+
+  function operate(callback, num) {
+    callback(num);
+
+    return calculator;
+  }
```
jstmpelowycz, line 68: let's put this function to object? no need to declare them above and then paste them below

jstmpelowycz (changes requested):

PR #1404, opened 2023-03-27
Solution

Zibi95:
I have problem with 'this' in operation methods using as callback in operate method. Couldn't resolve that, so instead i used closures.

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,36 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  return {
+    result: result,
+    add(a) {
+      result += a;
+    },
+    subtract(a) {
+      result -= a;
+    },
+    divide(a) {
+      result /= a;
+    },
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
```
Radoslaw-Czerniawski, line 40: This works however you are creating a [closure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures), changing variable inside and assigning it's value to the actual property on the object. You could try a different approach using [call method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) in the `operate` method while calling the `operation` parameter. Then you could change value of `result` property directly in object's methods.

Radoslaw-Czerniawski (approved):
Good solution, pr accepted, however I would like to see one with the `.call` method. In some spare time try to implement it :)

PR #1406, opened 2023-03-27
solution

DanielSorokowski:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    add(number, obj) {
```
solaryasha, line 41: It's strange looking that these `add` `subtract` and so on methods take as an argument some `obj`
Let's will do that these methods will take **only** the number

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      callback(num);
+
+      return this;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+    result: 0,
```
solaryasha, line 63: You could use closure and have this result inside the body function, not in the object.

solaryasha (approved):
You could play with `this` keyword instead of `calculator`. But it's fine!

PR #1407, opened 2023-03-27
Solution

Vasyl-Zhyliakov:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+      result *= x;
+    },
+    divide(x) {
+      result /= x;
+    },
+    reset() {
+      result = 0;
+
+      return calculator;
+    },
```
loralevitska, line 59: Add empty line between methods 
```suggestion
    add(x) {
      result += x;
    },
    
    subtract(x) {
      result -= x;
    },
    
    multiply(x) {
      result *= x;
    },
    
    divide(x) {
      result /= x;
    },
    
    reset() {
      result = 0;

      return calculator;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+    operate(callback, value) {
+      callback(value);
```
loralevitska, line 62: Let's more practice with `this`
```suggestion
      callback.call(this, value);
      
      return this
```

loralevitska (changes requested):
Almost done) Let's improve

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
+    reset() {
+      result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, value) {
+      callback.call(this, value);
+
+      return calculator;
```
DarkMistyRoom, line 68:
```suggestion
      return this;
    },

    operate(callback, value) {
      callback.call(this, value);

      return this;
```

DarkMistyRoom: it should work with any entity of this calculator

DarkMistyRoom: I suggest you make an experiment: try to create two calculators and work with them. ask in the chat if you have questions

DarkMistyRoom (changes requested):

PR #1413, opened 2023-03-30
Solution

JaneShavrukova:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    add(num1, num2) {
+      return num1 + num2;
+    },
```
tetiana-ivanchenko, line 43: image

tetiana-ivanchenko (changes requested):

src/makeCalculator.js
JaneShavrukova: I implemented this in `operate`  function. If I set value in the `add` function, the callback has no sense.

tetiana-ivanchenko: sorry, it's my fault, your solution is OK

tetiana-ivanchenko (approved):

PR #1419, opened 2023-04-02
solution

MikiOOOO:


Dka888 (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(resultSoFar, number) {
+      return resultSoFar + number;
+    },
```
d-wasilewski, line 50: this should take only number as an argument and assign the result to the `result` variable

d-wasilewski (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+    divide(number) {
+      this.result /= number;
+    },
+    multiply(number) {
+      this.result *= number;
+    },
+
+    operate(operation, number) {
+      operation.bind(this, number)();
```
d-wasilewski, line 62: It is less clear than simply calling the operation method directly. Additionally, the bind method creates a new function object, which could lead to performance issues if called repeatedly

d-wasilewski (changes requested):

src/makeCalculator.js
```diff
@@ -41,27 +41,27 @@ function makeCalculator() {
...
-      this.result *= number;
+      calculator.result *= number;
     },
 
     operate(operation, number) {
-      operation.bind(this, number)();
+      operation(number);
 
-      return this;
+      return calculator;
```
d-wasilewski: you can still return this here, that was fine

d-wasilewski (changes requested):

src/makeCalculator.js
MikiOOOO: yes I know, but I used ctrl + d to target all "this" occurances and that is why it got changed as well. Is it neccessary to switch it back to this? Isn't it basically the same?

d-wasilewski: It's true, returning either `calculator` or `this` from the makeCalculator function will both return the same object. One small thing is that if you change the name in the future you won't have to refactor it in all places, `this` will always refer to the same.

There is also a thing thqt returning `calculator` and `this` is different in the way the object is accessed after the makeCalculator function is called. If `this` is returned, the object will be accessed through the variable or context in which the function was called

It's a subtle difference but it's better to keep it that way.

MikiOOOO: okay :D

d-wasilewski (approved):

PR #1427, opened 2023-04-05
solution

NikaNika12:


src/makeCalculator.js
```diff
@@ -38,6 +38,48 @@
  */
 function makeCalculator() {
   // write code here
-}
+  const calculator = {
+
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
```
etojeDenys, line 46:
```suggestion
      this.result += number;
```

etojeDenys (changes requested):
we need to use `this` instead of `calculator`

src/makeCalculator.js
NikaNika12: тогда вылазят ошибки и не видит this

      56 |
      57 |     subtract(number) {
    > 58 |       this.result -= number;
         |       ^
      59 |

tetiana-ivanchenko (approved):

PR #1429, opened 2023-04-06
first solution

spencerspe:


src/makeCalculator.js
```diff
@@ -38,6 +38,41 @@
...
+    },
+
+    operate(callback, aNumber) {
+      callback(aNumber);
+
+      return this;
+    },
+
+    add(a) {
+      calculator.result = calculator.result + a;
```
Anastasiia-Svintsova, line 59: You can shorten it. Same goes to all methods
```suggestion
      calculator.result += a;
```

src/makeCalculator.js
```diff
@@ -38,6 +38,41 @@
  */
 function makeCalculator() {
   // write code here
+
+  const calculator = {
+
+    result: 0,
```
Anastasiia-Svintsova, line 44: Don't add redundant spaces
```suggestion
  const calculator = {
    result: 0,
```

Anastasiia-Svintsova (approved):

PR #1430, opened 2023-04-06
task makeCalc done

HYPER2307:


src/makeCalculator.js
```diff
@@ -1,43 +1,32 @@
...
+    },
+    operate(callback, num) {
+      calc.result = callback(num);
+      return this;
+    },
+    reset() {
+      calc.result = 0;
+      return this;
+    }
+  };
```
IvanFesenko, line 27: Refactor. You shoud work with this inside methods instead of `calc.result`.
Expected result: `this.result`

IvanFesenko (changes requested):

src/makeCalculator.js
HYPER2307: при використанні this пише що неможливо прочитати властивість result у undefined
image

src/makeCalculator.js
```diff
@@ -1,43 +1,33 @@
...
+      this.result -= a;
+    },
+    multiply(a) {
+      this.result *= a;
+    },
+    divide(a) {
+      this.result /= a;
+    },
+    operate(callback, num) {
+      callback.call(this, num);
```
mykhalenych, line 20: The `operate` method should call the callback method using the calculator object's context. Instead of `callback.call(this, num)`, use `callback.call(calc, num)`.

mykhalenych (changes requested):

src/makeCalculator.js
```diff
@@ -1,43 +1,34 @@
...
+      this.result -= a;
+    },
+    multiply(a) {
+      this.result *= a;
+    },
+    divide(a) {
+      this.result /= a;
+    },
+    operate(callback, num) {
+      callback.call(calc, num);
```
etojeDenys, line 20: Use `this` instead of `calc` to maintain consistency in the code: callback.call(this, num);

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #1433, opened 2023-04-06
Solution

Tarasova09:


src/makeCalculator.js
```diff
@@ -36,8 +36,42 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
   // write code here
+  const calculator = {
```
etojeDenys, line 42: Use `this` instead of referencing `calculator` within the object methods. This will make the code more flexible and maintainable. For example, change `calculator.result += num;` to `this.result += num;`.

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #1434, opened 2023-04-06
Solution

vlad-liakh:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (value) => {
+      calculator.result += value;
```
etojeDenys, line 44: Use `this` instead of referencing calculator within the object methods. This will make the code more flexible and maintainable. For example, change `calculator.result += num;` to `this.result += num;`.

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1435, opened 2023-04-09
Solution

AnastassiaH:


src/makeCalculator.js
```diff
@@ -38,6 +38,45 @@
...
+
+      return this;
+    },
+    divide(value) {
+      this.result /= value;
+
+      return this;
+    },
+
+    reset(value) {
```
nazarmatsevych, line 66: The 'reset' method should not have a parameter 'value' as it's not used. Please remove it. 🧹

nazarmatsevych (approved):
Good work,
just small suggestion

PR #1439, opened 2023-04-12
solution

mariuszgit:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+
+      return this;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+    operate(callback, number) {
+      callback(number);
```
DorotaLeniecDev, line 68: Use 'this' as the context for the callback function by calling 'callback.call(this, number)'

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(num) {
+      calculator.result += num;
```
DorotaLeniecDev, line 43: Use 'this.result' instead of 'calculator.result' to reference the current object

DorotaLeniecDev (changes requested):
Check my comments :)

DorotaLeniecDev (approved):
Good job :)

PR #1442, opened 2023-04-15
Solution

andrijdudar:


src/makeCalculator.js
```diff
@@ -33,11 +33,41 @@
...
- * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(x) {
+      calculator.result += x;
+    },
```
SanyaBratashchuk, line 43: Change result in operate and reset methods only 
```suggestion
    add(a, b) {
      return a + b;
    },
```

SanyaBratashchuk (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
Great!

PR #1444, opened 2023-04-17
1.0.0 all tests passed solution

domkab:


src/makeCalculator.js
```diff
@@ -1,43 +1,49 @@
...
+  };
+
+  const operate = (callback, num) => {
+    callback(num);
+
+    return innerCalculator;
+  };
+
+  const innerCalculator = {
+    add,
```
anstsot, line 35: add result as a param of object
```suggestion
  const innerCalculator = {
    result: 0,
    add,
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

anstsot (approved):
Great job, approved!

PR #1448, opened 2023-04-21
make calculator

vladimirshub:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    divide(num) {
+      calculator.result /= num;
+    },
+
+    multiply(num) {
+      calculator.result *= num;
+    },
+
+    reset() {
+      calculator.result = 0;
```
IvanFesenko, line 60: You should use `this` instead of `calculator`

IvanFesenko (changes requested):
Some changes required

PR #1450, opened 2023-04-22
Solution

MTS13:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    operate(callback, num) {
+      callback.call(this, num);
+
+      return this;
+    },
+
+    add(num) {
+      this.result += num;
+
+      return this.result;
```
SanyaBratashchuk, line 58: change this.result in operate and reset methods only  
```suggestion
      return this.result + num;
```

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1463, opened 2023-04-30
Add solution

ruslanpashkov:


generatedman (commented):
good job!

generatedman (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```
naziamoff, line 57: you should not expect two parameters here, you can reference result from the inside of the method

naziamoff: as well as you can change this.result right here, instead of returning the variable and reassigning this.result in operate function

naziamoff (changes requested):
Not bad. A couple of things left and we're good 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
+    },
```
naziamoff, line 45: as you're inside the `calculator` here - you should reference its result directly from `this`
```suggestion
    add(number) {
      this.result += number;
    },
```

naziamoff (approved):
good 👍 
But you should use `this` if you are inside the method and trying to reference the object you're currently in

PR #1464, opened 2023-05-01
Solution

generatedman:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+      return res - val;
+    },
+
+    multiply(val, res) {
+      return res * val;
+    },
+
+    divide(val, res) {
+      return res / val;
+    },
```
DidenkoDenis306, line 56: You don't need to expect two parameters here. You can reference the result from inside the method.

DidenkoDenis306 (approved):
Great!

33Orange (changes requested):
Well done!

You don't need to return `this` inside all methods. Only `operate` and `reset` can be chained. 
Everything else is well done

dinaminahi (approved):
👍

PR #1475, opened 2023-05-05
Solution

vovaAlieksieiev:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    add,
+    subtract,
+    multiply,
+    divide,
+    reset,
+    operate,
+    get result() {
+      return result;
+    },
+  };
```
Volodymyr-Melnyk-Fron10der, line 80: move all methods and variables into an object and return the object to be consistent with the task

Volodymyr-Melnyk-Fron10der (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
great!

PR #1479, opened 2023-05-08
add makeCalculator solution

DidenkoDenis306:


ruslanpashkov (approved):

Anna-Tkachenko (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num1, num2) {
+      return num1 + num2;
+    },
+
+    subtract(num1, num2) {
```
33Orange, line 47: The subtract function should only take one parameter and subtract it from the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num1, num2) {
```
33Orange, line 43: The add function should only take one parameter and add it to the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    subtract(num1, num2) {
+      return num1 - num2;
+    },
+
+    multiply(num1, num2) {
+      return num1 * num2;
+    },
+
+    divide(num1, num2) {
```
33Orange, line 55: The divide function should only take one parameter and divide the current result by it.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    add(num1, num2) {
+      return num1 + num2;
+    },
+
+    subtract(num1, num2) {
+      return num1 - num2;
+    },
+
+    multiply(num1, num2) {
```
33Orange, line 51: The multiply function should only take one parameter and multiply the current result with it.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, n) {
+      this.result = operation(this.result, n);
```
33Orange, line 66: Use `operation.call(context, argument)` to call function with new context and argument.

33Orange (changes requested):
A few moments:
- Update calculate methods. Functions should only take one parameter and add it to the current result.
- When you update methods use `.call()` to call your callback with provided context and arguments.

dinaminahi (changes requested):
👍

naziamoff (approved):
🔥

PR #1480, opened 2023-05-08
Solution

AlexRaven93:


src/makeCalculator.js
```diff
@@ -38,6 +38,33 @@
...
   // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, b) {
+      callback(b);
+
+      return this;
+    },
+    add: (b) => {
+      calculator.result += b;
```
SanyaBratashchuk, line 49: change result  in operate method only

SanyaBratashchuk, line 49:
```suggestion
      this.result += b;
```
Make it work with this

SanyaBratashchuk (changes requested):

DarkMistyRoom (changes requested):
not fixed
image
 If I call `add` method directly, it will affect the result

anstsot (approved):
Great job, approved!

PR #1483, opened 2023-05-09
Advanced calculator solution

ZupaFly:


qoa11a (changes requested):
Your `add`, `subtract`, `multiply`, and `divide` methods don't have to return `this`.

PR #1489, opened 2023-05-10
Solution

VolodymyrPst:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
-  // write code here
+  const obj = {
+    result: 0,
+
+    operate(callback, operand) {
+      return callback(operand);
+    },
+
+    reset() {
+      obj.result = 0;
```
slavanosatskyi, line 48: it should be 'this' everywhere instead of obj. Technically this will be equal to obj, but still it is less confusing

slavanosatskyi (changes requested):

PR #1491, opened 2023-05-10
add task solution

maksym-nemera:


src/makeCalculator.js
```diff
@@ -37,7 +37,55 @@
...
+      return this;
+    },
+
+    get add() {
+      return (value) => {
+        this.result += value;
+
+        return this.result;
+      };
+    },
```
ruslanpashkov, line 55: are you sure that our methods should be getters?

src/makeCalculator.js
```diff
@@ -37,7 +37,55 @@
...
+      callback(value);
+
+      return this;
+    },
+
+    get add() {
+      return (value) => {
+        this.result += value;
+
+        return this.result;
```
ruslanpashkov, line 53: we pass our data directly to our method, we don't need to create a closure here for this

ruslanpashkov (changes requested):

silchyn (approved):

PR #1492, opened 2023-05-10
Solution

BogdanParubchenk01:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+
+    add(value) {
+      obj.result += value;
```
Viktor-Kost, line 44: apply for other methods as well
```suggestion
      this.result += value;
      return this
```

Viktor-Kost (changes requested):

VitaliyBondarenko1982 (approved):

PR #1493, opened 2023-05-10
Solution

Pa1eOrc:


src/makeCalculator.js
```diff
@@ -38,6 +38,38 @@
  */
 function makeCalculator() {
   // write code here
```
OleziO, line 40: Remove all comments from the code

OleziO (approved):

PR #1496, opened 2023-05-11
solution

skinnyjazz:


33Orange (changes requested):
Well done ! 🔥 

Just a one moment with readability:
- Your function return nameless object. Use proper names for variables. If function was called `makeCalculator`, it should return `calculator`. This will improve the readability of the code.

dinaminahi (approved):
Well done👍

PR #1503, opened 2023-05-14
add functionality to makeCalculator

cephy77:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      that.result -= number;
+    },
+
+    multiply(number, that) {
+      that.result *= number;
+    },
+
+    divide(number, that) {
+      that.result /= number;
+    },
```
DidenkoDenis306, line 69: Functions should only take one parameter and add it to the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
```
DidenkoDenis306, line 40: Would be better to create an calculator object and return it

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    result: 0,
+
+    operate(opetation, number) {
```
DidenkoDenis306, line 43: Let`s rename first parametr to "operation". And use .call() to call your callback with provided context and arguments.

DidenkoDenis306 (changes requested):
Good job, but it would be cool to fix a few points)

src/makeCalculator.js
cephy77: Oh, nice! I didn't know about this method)

src/makeCalculator.js
cephy77: I had made it to bind the context, but the call method fix this issue

src/makeCalculator.js
cephy77: Actually, I see no point to that action

dinaminahi (approved):

naziamoff (approved):

PR #1504, opened 2023-05-14
calculator

diana-tuz:


src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  let result = 0;
```
vlad-demchuk, line 40: result variable should be a part of object, not just in the function

vlad-demchuk (changes requested):

PR #1507, opened 2023-05-15
Solution

julshvets:


src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
```
33Orange, line 44: The subtract function should only take one parameter and subtract it from the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (a, b) => a + b,
```
33Orange, line 43: The add function should only take one parameter and add it to the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
...
+    subtract: (a, b) => a - b,
+    divide: (a, b) => a / b,
+    multiply: (a, b) => a * b,
```
33Orange, line 46: The multiply function should only take one parameter and multiply the current result with it.

src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    divide: (a, b) => a / b,
```
33Orange, line 45: The divide function should only take one parameter and divide the current result by it.

src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
...
+    multiply: (a, b) => a * b,
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(operation, number) {
+      this.result = operation(this.result, number);
```
33Orange, line 53: Use `operation.call(context, argument)` to call function with new context and argument.

33Orange (changes requested):
Good job.

A few moments:

- Update calculate methods. Functions should only take one parameter and add it to the current result.
- When you update methods use .call() to call your callback with provided context and arguments.

33Orange (changes requested):
Well done ! 🔥 

A few moments with readability:
- Lets add 1 space between methods.
- Rename parameter `a` to something more descriptive. For example: `num`
- You don't need to `return this` in all methods. Only `operate` and `reset` 
can be used in a chain.

dinaminahi (approved):
Well done 👍

PR #1509, opened 2023-05-16
made task ja advanced calculator

vladyslav-lapshyn:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+    result = 0;
+
+    return calc;
+  };
+
+  const operate = (operation, number) => {
+    operation(number);
+
+    return calc;
+  };
```
dinaminahi, line 68: What if I make 2 objects with makeCalculator() and pass the method of second one to the `operate` of first one?
image

I think that it is suppose to change the object on which `operate` was called. Think of using methods that are changing the context  of the invoking function.

dinaminahi (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+
+  const calc = {
```
dinaminahi, line 70: Rename it to `calculator`. It's better not to shorten variable names

src/makeCalculator.js
vladyslav-lapshyn: @dinaminahi honestly, I couldn't understand how can I do it else, only like rewrite solution and put all that in 1 common object. Now it works, if I based on your example

33Orange (approved):
Well done ! 🔥

PR #1511, opened 2023-05-16
makeCalculator solution

PANCH1811:


src/makeCalculator.js
```diff
@@ -37,7 +37,55 @@
...
+  }
+
+  function reset() {
+    result = 0;
+
+    return this;
+  }
+
+  function operate(callback, num) {
+    callback(num);
```
33Orange, line 73: Use .call() or .apply() to explicitly set the context of 'this' for the callback function.

33Orange (changes requested):
Well done !

Some comments:
- Let's rewrite the solution using object `methods` and `this`.
- Here's what happens if I pass the method of the second calculator into the operate method of the first

image

- Only `reset` and `operate` must return this for chaining.

dinaminahi (approved):
Good job 👍

PR #1514, opened 2023-05-16
calculator

DenisGurskiy:


src/makeCalculator.js
```diff
@@ -38,6 +38,35 @@
...
+  const calculator = {
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(a) {
+      calculator.result += a;
```
Viktor-Kost, line 50: use 'this' instead of 'calculator' inside the methods. Replace 'calculator.result' with 'this.result'. Apply for other cases as well

Viktor-Kost (changes requested):
Good job!
Just minor fixes needed

src/makeCalculator.js
DenisGurskiy: With 'this.result' tests didn't work. I know how do that work (I saw this in fe_chat4) but I don't fully understand how this work. I must use calculator.call(this, number).

src/makeCalculator.js
```diff
@@ -38,6 +38,35 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(a) {
+      this.result += a;
+    },
```
Viktor-Kost, line 51:
```suggestion
    add(a) {
      this.result += a;
      
      return this;
    },
```

Viktor-Kost (changes requested):
Almost there. 
`return this;` is added not to all methods

src/makeCalculator.js
DenisGurskiy: The task says that only the 'operate' and 'reset' methods are called by the chain. return.this don't use in another methods

vlad-demchuk (approved):

PR #1516, opened 2023-05-18
add solution

niaYaro:


naziamoff (changes requested):
let's fix linting errors instead of disabling them 😄

33Orange (approved):
Well done ! 🔥

naziamoff (approved):

PR #1521, opened 2023-05-20
add task solution

KatyaZaremska:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const objCalc = {
```
33Orange, line 40:
```suggestion
  const calculator = {
```
Would be better. Adding the word `obj` does not improve understanding of what this variable is.

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    operate(callback, num) {
+      callback.call(this, num);
+
+      return this;
+    },
+
+    add(num = 0) {
+      this.result += num;
+
+      return this;
```
33Orange, line 52: By the condition of the assignment, only the `operate` and `reset` functions return an object, to create a chain of calls

33Orange (changes requested):
Well done ! 🔥 

A few moments: 
- Rename object which you return in `makeCalculator`.
- Remove `return this` from `add`, `subtract`, `multiply`, `divide`

33Orange (approved):
Awesome ! ⭐

PR #1523, opened 2023-05-22
Solution

maryanmats:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add: function(num) {
```
silchyn, line 42:
```suggestion
    add(num) {
```

silchyn, line 42:
```suggestion
    result: 0,
    
    add: function(num) {
```

silchyn (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      this.result += num;
+
+      return this;
+    },
```
33Orange, line 47: By the condition of the task, only operate and reset must return this

```suggestion
    add(num) {
      this.result += num;
    },
```

33Orange (approved):
Well done !

PR #1527, opened 2023-05-22
code

liza-strykharchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    },
+
+    multiply(num) {
+      this.result *= num;
+
+      return this;
+    },
+
+    divide(num) {
+      this.result /= num;
```
volodymyr-soltys97, line 62: Add a check for division by zero in the 'divide' function to prevent unexpected results.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+      return this;
+    },
+
+    reset(num) {
```
volodymyr-soltys97, line 67: The 'reset' function should not have any parameter, as it is not used in the function body.

volodymyr-soltys97 (changes requested):
Good job 👍 

slavanosatskyi (approved):

PR #1534, opened 2023-05-28
Solution

LiubomyrShyshak:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    multiply: function(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide: function(num) {
+      this.result /= num;
+
+      return this;
+    },
```
qoa11a, line 61: These methods shouldn't return this. Only operate and reset should.

qoa11a (approved):

qoa11a (approved):

polosanya (approved):
There is no need to re-request the review, this task is already approved 🙂

PR #1544, opened 2023-06-01
solution

vlad-fil:


src/makeCalculator.js
```diff
@@ -38,6 +38,34 @@
...
+    },
+    subtract(number) {
+      return calculator.result - number;
+    },
+    divide(number) {
+      return calculator.result / number;
+    },
+    multiply(number) {
+      return calculator.result * number;
+    },
```
etojeDenys, line 65: let's use `this.  ...` in methods instead of `calculator`

(and don't forget to bind your callback)

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1545, opened 2023-06-02
make calculator

w315:


src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+    result: 0,
+
+    add(number) {
+      const operation = calculator.result + number;
+
+      return operation;
+    },
+
+    subtract(number) {
+      const operation = calculator.result - number;
```
volodymyr-soltys97, line 51: Instead of creating a new variable 'operation', directly update the 'result' property

volodymyr-soltys97 (changes requested):
Good job 👍 

VitaliyBondarenko1982 (approved):

PR #1546, opened 2023-06-02
Solution

andrii-rykal:


src/makeCalculator.js
```diff
@@ -32,12 +32,42 @@
...
- *
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      return calculator.result + number;
```
VitaliyBondarenko1982, line 42: here and in other methods better to use this keyword
```suggestion
      return this.result + number;
```

VitaliyBondarenko1982 (approved):
Well done

PR #1549, opened 2023-06-09
Solution

AllaSerhiienko:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(a) {
+      return calculator.result + a;
```
SanyaBratashchuk, line 54:
```suggestion
      return calculator.result + a;
```
replace calculator with this everywhere

SanyaBratashchuk (changes requested):

TarasHoliuk (approved):

PR #1550, opened 2023-06-09
Solution

dmytro-popovych:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num) {
+      calculator.result += num;
```
SanyaBratashchuk, line 44: replace calculator with this everywhere

SanyaBratashchuk (changes requested):

src/makeCalculator.js
dmytro-popovych: when i did it the tests fail

dmytro-popovych: I can't understand why

`function makeCalculator() {
  const calculator = {
    result: 0,

    add(num) {
      this.result += num;

      return calculator;
    },

    subtract(num) {
      this.result -= num;

      return calculator;
    },

    multiply(num) {
      this.result *= num;

      return calculator;
    },

    divide(num) {
      this.result /= num;

      return calculator;
    },

    reset() {
      this.result = 0;

      return calculator;
    },

    operate(callback, num) {
      callback(num);

      return calculator;
    },
  };

  return calculator;
}`

dmytro-popovych: I did it, but I still don't understand why a simple replacement broke everything.

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, num) {
+      callback.call(this, num);
```
TarasHoliuk, line 74: about your question

that's (`call` method) fixed your code because when you just call `callback` which `operate` method accepts in arguments that's calls as a common function and don't know what is `this` in your callback

this is known if you call a method in this way, right from an object: calculator.add, etc.

TarasHoliuk (approved):

PR #1552, opened 2023-06-11
add solution

prostonadobobb:


src/makeCalculator.js
```diff
@@ -37,7 +37,28 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (x, y) => x + y,
```
anastasiiavorobiova, line 43: Add method should only take one argument and add it to the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,28 @@
...
+    subtract: (x, y) => x - y,
```
anastasiiavorobiova, line 44: Subtract method should only take one argument and subtract it from the current result.

src/makeCalculator.js
```diff
@@ -37,7 +37,28 @@
...
+    multiply: (x, y) => x * y,
+    divide: (x, y) => x / y,
```
anastasiiavorobiova, line 46: Divide method should only take one argument and divide the current result by it.

anastasiiavorobiova (changes requested):
Great start!

nazarmatsevych (changes requested):
Hey, looks like you didn't push anything since last review.

If you have any questions - feel free to ask them in fe_chat

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    subtract(number) {
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      this.result /= number;
```
volodymyr-soltys97, line 57: Add a condition to check if the number is not zero before dividing to avoid division by zero.

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    add(number) {
+      this.result += number;
+    },
+
+    subtract(number) {
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
```
volodymyr-soltys97, line 53: Consider returning the updated result after each operation for better usability.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/makeCalculator.js
```diff
@@ -38,6 +38,42 @@
...
+    },
+
+    divide(number) {
+      if (number === 0) {
+        return 'error';
```
volodymyr-soltys97, line 64: Try to create a more descriptive message
```suggestion
        return 'Can not divide by 0, try another number';
```

volodymyr-soltys97 (approved):
Nice work 🚀

PR #1553, opened 2023-06-12
solution

SergeyGeorg:


src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const object = {};
```
witflash, line 41: It's better to name it `calculator` instead of a general `object` name.

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+
+  object.multiply = function multiply(value) {
+    object.result = object.result * value;
+  };
+
+  object.reset = function reset() {
+    object.result = 0;
+
+    return object;
+  };
```
witflash, line 71: You can create all methods directly inside your `calculator` object. The next way:
```
const calculator = {
  result: 0,

  operate(action, value) {
    ...
  }
  
  add(value) {
    ...
  }
  
  // etc. 
}
```

So, you will have shorter and cleaner code. 
You can try this approach.

witflash (changes requested):
Hi!
Your solution is workable, but you can improve your code a little bit. 
Check my comments, please.

PR #1556, opened 2023-06-12
solution

maxyma-sanx:


KandaskalovOleg (approved):
Approve ✔️

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    divide(value) {
+      calculator.result /= value;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
```
tetiana-ivanchenko, line 62: why for operation you use calculation, but return this?

tetiana-ivanchenko (changes requested):

YegorKorenkov (approved):

PR #1558, opened 2023-06-12
task solution

Bonum-taurus:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      this.method = callback;
```
volodymyr-soltys97, line 44: You should bind the callback to the calculator object to ensure 'this' is properly set when calling the method. Replace 'this.method = callback;' with 'this.method = callback.bind(this);'

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+
+      this.method(number);
```
volodymyr-soltys97, line 46: You don't need to store the method in the calculator object. You can directly call the bound callback. Replace 'this.method(number);' with 'callback.call(this, number);'

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      this.result /= number;
+    },
```
volodymyr-soltys97, line 65: Add a check for the number, since we cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 

mykhalenych (approved):

PR #1560, opened 2023-06-12
add task solution

ilona-borkovska:


KandaskalovOleg (approved):
Good work ✨

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, num) {
+      this.result = operation(this.result, num);
```
anstsot, line 66: you can avoid propagating `this.result` in `operation` here and use directly `this.result` in each method (`add`, `subtract` etc.)

PR #1569, opened 2023-06-14
solution

cebanoleksandr:


src/makeCalculator.js
```diff
@@ -37,7 +37,32 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
```
MariiaMonakhova, line 40: [checklist: 1] if you have a function that is called makeCalculator, it should return calculator. Instead of return make a calculator object and return it at the end.

MariiaMonakhova (changes requested):

anstsot (approved):
Good work, approving:sparkles:

PR #1576, opened 2023-06-15
solution

MarianVolytskyi:


ivanbilotserkivskiy (approved):
GJ! 👍

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    multiply(value) {
+      newCalculator.result *= value;
+    },
+
+    divide(value) {
+      newCalculator.result /= value;
+    },
+
+    reset() {
+      this.result = 0;
```
tetiana-ivanchenko, line 61: why do you use `newCalculator.result` in all others methods, but `this.result` here?

tetiana-ivanchenko (approved):

PR #1577, opened 2023-06-15
calculator solution

ivanbilotserkivskiy:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+    multiply(value) {
+      calculator.result *= value;
+    },
+
+    divide(value) {
+      calculator.result /= value;
+    },
+
+    reset() {
+      this.result = 0;
```
tetiana-ivanchenko, line 65: why do you use `calculator.result` in the others methos, but `this.result` here?

tetiana-ivanchenko (approved):

src/makeCalculator.js
ivanbilotserkivskiy: I had to rewrite code and I forgot to rewrite it there)

PR #1579, opened 2023-06-15
SolutionCalculator

anastasiia-burkovska:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num) {
+      calculator.result += num;
```
Viktor-Kost, line 44: Instead of directly using 'calculator.result', use 'this.result' to access the current object's property.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    divide(num) {
+      calculator.result /= num;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
```
Viktor-Kost, line 62:
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    operate(call, num) {
+      call(num);
+
+      return calculator;
```
Viktor-Kost, line 68:
```suggestion
      return this;
```

Viktor-Kost (changes requested):
Nice!
Some moments can be improved

VitaliyBondarenko1982 (approved):

PR #1581, opened 2023-06-15
Solution

adv555:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    multiply(num) {
+      calculator.result *= num;
+    },
+
+    divide(num) {
+      calculator.result /= num;
+    },
+
+    reset() {
+      this.result = 0;
```
tetiana-ivanchenko, line 60: why do you use `calculator.result` in the others methos, but `this.result` here?

tetiana-ivanchenko (approved):

PR #1582, opened 2023-06-16
solution

GGsmou:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
```
outofs, line 40: GW! But i'm not sure about first item in checklist:
1. [NAMING] - Use proper names for variables. **If function was called `makeFruit`, it should return `fruit`**

But I think it is not critical, you just returned proper value

outofs (approved):
Approve!

PR #1583, opened 2023-06-16
Solution

MelnychukVol:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add: function(num) {
```
agniya-i, line 42:
```suggestion
    add(num) {
```
if you creating a method in the object, you don't need to use function keyword, use shortcut instead 👍

agniya-i (approved):

PR #1584, opened 2023-06-16
add solution

akozlovska:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      return result + number;
+    },
+    subtract(number, result) {
+      return result - number;
+    },
+    multiply(number, result) {
+      return result * number;
+    },
+    divide(number, result) {
+      return result / number;
```
Andriy-Shymkiv, line 62: what if `result === 0`? you can't divide by zero, so you should probably throw an error in that case

Andriy-Shymkiv (changes requested):

src/makeCalculator.js
VitaliyBodnar: not fixed, because the result can also be a '0'

VitaliyBodnar (changes requested):

akozlovska:
Thanks for pointing this out! Now it is fixed. Thought that only division by zero (if number is 0) is prohibited, but if the result is 0, it can be divided by any number and the result of the division will be 0, it doesn't contradict mathematical rules

TarasHoliuk (approved):

PR #1587, opened 2023-06-18
Solution

olena-ole:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    operate(foo, number) {
```
GGsmou, line 43: I would place operate under other operations to be consistent, but it's not a big deal)
Name `foo` is not really descriptive, consider renaming `foo` to `callback` or `operation`

GGsmou (changes requested):
GJ
Please fix some issues)

GGsmou (approved):
GJ!
Approving)

YegorKorenkov (approved):
GJ!

PR #1591, opened 2023-06-18
Solution

yurii-krymskiy:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: (number) => {
+      calculator.result += number;
```
mykhalenych, line 43: Instead of using 'calculator.result', use 'this.result' to reference the object property. Check all methods

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: (number) => {
```
mykhalenych, line 42: To properly bind 'this' context in the methods, change arrow functions to regular functions.

src/makeCalculator.js
OleziO: Not resolved.

OleziO (changes requested):

PR #1593, opened 2023-06-19
calculator

OleksiiBielovolov:


src/makeCalculator.js
```diff
@@ -37,7 +37,66 @@
...
+  }
+
+  function divideNumber(number) {
+    calculator.result /= number;
+  }
+
+  function resetResult() {
+    calculator.result = 0;
+
+    return this;
```
volodymyr-soltys97, line 96: Use 'return this;' instead of 'return calculator;' to maintain the correct context.

PR #1599, opened 2023-06-19
Solution

dipinchuk:


artkas03 (approved):

YegorKorenkov (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      this.result += value;
+
+      return this;
```
YegorKorenkov, line 46: Return this in each operation method not required, you already return this in **operate** method

YegorKorenkov (approved):
GJ!

PR #1600, opened 2023-06-19
Solution

kylyvnyk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
```
dianadiv, line 46: you can use keyword 'this' instead of 'calculator'

dianadiv (approved):
Good job, approved✅

src/makeCalculator.js
YegorKorenkov, line 46: Use 'return this;' instead of 'return calculator;' to maintain the correct context. Please check all methods

YegorKorenkov (changes requested):

YegorKorenkov (approved):

PR #1607, opened 2023-06-22
Make calculator task

ValeraViachalo:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+    firstOperator + secondOperator;
+
+  calculator.subtract = (firstOperator, secondOperator) =>
+    firstOperator - secondOperator;
+
+  calculator.multiply = (firstOperator, secondOperator) =>
+    firstOperator * secondOperator;
+
+  calculator.divide = (firstOperator, secondOperator) =>
+    firstOperator / secondOperator;
```
Viktor-Kost, line 66: your calculation methods should look like this
```
    add(value) {
      this.result += value;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, value) {
+      this.result = operation(this.result, value);
+
+      return this;
+    },
```
Viktor-Kost, line 53: here you can use like this
```
    operate(operation, value) {
      operation.call(this, value);

      return this;
    },
```

Viktor-Kost (changes requested):

VitaliyBondarenko1982 (approved):

PR #1608, opened 2023-06-24
Solution

Luk2asz:


Radoslaw-Czerniawski (approved):
Good job, works like it should :)
If you find some time you could try implementing this with immediately retuning an object instead of using a variable, it will require using `this` keyword and `call` method in `opearate`.

PR #1609, opened 2023-06-25
solution

707Andrew:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  const calculator = {
+    add: function(num) {
+      result += num;
+
+      return calculator;
+    },
```
SanyaBratashchuk, line 47:
```suggestion
    add(num) {
      result += num;

      return calculator;
    },
```
Fix it everywhere

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    subtract: function(num) {
+      result -= num;
+
+      return calculator;
+    },
```
SanyaBratashchuk, line 52: use this instead of calculator

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):
Interesting solution with getter so I approve it. But for better understanding of `this` try to rewrite with result as object property instead of external variable

PR #1611, opened 2023-06-27
add calculator

dubyuliia:


src/makeCalculator.js
```diff
@@ -32,12 +32,44 @@
...
  * calculator.result === -28
- *
+
  *
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
```
```suggestion
  return {
```
you might immediately return created object instead of creating a separate variable (FYI)

TarasHoliuk (approved):

PR #1613, opened 2023-06-28
Solution AB

AytajBabayeva:


lerastarynets (approved):
Well done! Short and clear, I like it)

PR #1614, opened 2023-06-28
add task advanced_calculator

P-Nazar:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+  };
+
+  calculator.add = function(number) {
+    calculator.result += number;
+  };
```
VitaliyBondarenko1982, line 46: better to make method fields directly in calculator object, and after that will be possible to use `this` keyword
```suggestion
  const calculator = {
    result: 0,
    
    add(number) {
       this.result += number;
    };
    
    // same for other methods
    
     operate(callback, number) {
        callback.call(this, number);

      return this;
    },
  };
```
and in this case use call method for calling callback in operate method

VitaliyBondarenko1982 (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #1616, opened 2023-06-28
added solution

Constantin-Haranskiy:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+      operation.call(this, a);
+
+      return this;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(a) {
```
agniya-i, line 52:
```suggestion
    },
    
    add(a) {
```

It would be better for readability to add empty line between two methods

PR #1617, opened 2023-06-29
add solution

purpleFade:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+  return {
+    result: 0,
+
+    add: function(num) {
+      this.result += num;
+    },
+
+    subtract: function(num) {
+      this.result -= num;
+    },
```
SerhiiKirik, line 49: you can do it just like that
```suggestion
    add(num) {
      this.result += num;
    },

    subtract(num) {
      this.result -= num;
    },
```

SerhiiKirik (approved):
Well done 🔥

PR #1618, opened 2023-06-29
mySolution

AntonFof:


witflash (approved):
Good job!

You can separate every method with an empty line. It will look better

PR #1619, opened 2023-06-30
Develop

AnnaViolin23:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    add(num) {
+      calculator.result += num;
+
+      return calculator;
+    },
```
Viktor-Kost, line 45: use `this` at your methods
```suggestion
    add(num) {
       this.result += num;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+    multiply(num) {
+      calculator.result *= num;
+
+      return calculator;
+    },
+
+    divide(num) {
+      calculator.result /= num;
```
Viktor-Kost, line 60: it'd be nice to add a condition that num shouldn't be a zero. In opposite case you may return an error

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+      return calculator;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
```
Viktor-Kost, line 68:
```suggestion
      this.result = 0;

      return this;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    operate(callback, num) {
+      callback(num);
+
+      return calculator;
```
Viktor-Kost, line 74:
```suggestion
      return this;
```

Viktor-Kost (changes requested):

src/makeCalculator.js
Viktor-Kost: use this.result = 0;

src/makeCalculator.js
Viktor-Kost: use this.result += num;

Viktor-Kost (changes requested):
Nice!
Please pay attention to unresolved conversations since not all fixes were provided

Viktor-Kost (changes requested):
Please check whether you push your commit or not, since it still no changes 
If you need any help - feel free to ask at fe_chat

Viktor-Kost (approved):

PR #1620, opened 2023-06-30
solution

venher-v:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
+
+      return this;
+    },
+
+    add(result, number) {
+      return result + number;
+    },
```
Viktor-Kost, line 51: no needs to pass result, just use this. Apply for other calculation methods as well
```suggestion
    add(number) {
      this.result += number;
    },
```

Viktor-Kost (changes requested):

loralevitska (approved):
Well done)

PR #1621, opened 2023-07-01
made a calculator

vetal-hovenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(callback, operand) {
+      callback(operand);
```
volodymyr-soltys97, line 66: The 'callback' function might potentially change the context, so it's safer to bind the current context ('this') to it. Use 'callback.call(this, operand)' instead of 'callback(operand)'.

volodymyr-soltys97 (approved):
Good job 👍

src/makeCalculator.js
vetal-hovenko: ok, thanks

PR #1625, opened 2023-07-03
Solution

Sirius9312:
Calculate Advanced Task

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,48 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
+    digit: 0,
+
+    add: (val) => {
+      calc.digit += val;
+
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
+      calc.digit += val;
+
+      return this;
+    },
+    subtract: (val) => {
+      calc.digit -= val;
+
+      return this;
+    },
+    multiply: (val) => {
```
nazarmatsevych, line 53:
```suggestion
    add: (val) => {
      calc.digit += val;

      return this;
    },
    
    subtract: (val) => {
      calc.digit -= val;

      return this;
    },
    
    multiply: (val) => {
```

Next time pls add empty line between methods

nazarmatsevych (approved):

PR #1627, opened 2023-07-04
add task solution

NikitaOrliak:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    result: 0,
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
+
+    add(number) {
+      object.result += number;
```
mykhalenych, line 50: Inside the methods 'add', 'subtract', 'multiply', 'divide', 'reset', you are referring to the 'object' explicitly. A better way to do this would be using 'this' keyword, which refers to the object that is executing the current method.

mykhalenych (changes requested):

src/makeCalculator.js
NikitaOrliak: I cant use 'this' because I used callback in 'operate' method, so 'this' in my methods uses callback's 'this', not object. I tried to use 'this', but got undefined instead of an object. Actually, I found a different way to solve this problem - use .bind(calculator) in tests, but I think it is a wrong way :)

volodymyr-soltys97: Still not fixed

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const object = {
+    result: 0,
+
+    operate(callback, number) {
+      callback(number);
```
volodymyr-soltys97, line 44: Try to rewrite like this `this.result = callback(this.result, number)`

volodymyr-soltys97 (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    result: 0,
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
+
+      return this;
+    },
+
+    add(tempResult, number) {
+      const result = tempResult + number;
```
mykhalenych, line 50:
```suggestion
      this.result += num;
```
check all methods

mykhalenych (changes requested):

src/makeCalculator.js
NikitaOrliak: this.result doesn't work, 'this' in methods became undefined because methods use in callback function

NikitaOrliak: image
image

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const object = {
+    result: 0,
+
+    operate(callback, number) {
+      callback(this, number);
```
VitaliyBondarenko1982, line 44:
```suggestion
      callback.call(this, number);
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+      return this;
+    },
+
+    add(obj, number) {
+      obj.result += number;
+    },
```
VitaliyBondarenko1982, line 51:
```suggestion
    add(number) {
      this.result += number;
      
      return this;
    },
```
same for subtract, multiply and divide methods.

VitaliyBondarenko1982 (changes requested):
Please check my comments.
It should help to fix prev comments about using `this` instead `obj`

src/makeCalculator.js
NikitaOrliak: Yeap, it works. Thank you. But I don't really understand how it works. What does this '.call' mean? And, btw, I can remove 'return this' in methods and it still works.

DarkMistyRoom: You will learn `call` method later. `return this` is needed to call methods in a chain

DarkMistyRoom (approved):

PR #1628, opened 2023-07-04
Solution

misha-shevchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,25 @@
...
+    result: 0,
+    operate(func, num) {
+      this.result = func(this.result, num);
+
+      return this;
+    },
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => a / b,
```
volodymyr-soltys97, line 50: The 'divide' method does not handle the case when 'b' is zero. Dividing by zero in JavaScript will return Infinity, which may not be the expected behaviour.

src/makeCalculator.js
```diff
@@ -37,7 +37,25 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(func, num) {
+      this.result = func(this.result, num);
+
+      return this;
+    },
+    add: (a, b) => a + b,
```
volodymyr-soltys97, line 47: The 'add', 'subtract', 'multiply' and 'divide' methods are currently implemented as arrow functions. This means they don't have their own 'this' context, which may lead to unexpected behaviour if they are used within the 'operate' method. Consider changing them to regular functions.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      return a + b;
+    },
+    subtract(a, b) {
+      return a - b;
+    },
+    multiply(a, b) {
+      return a * b;
+    },
+    divide(a, b) {
+      return b === 0 ? 'ERROR! You can`t divide by zero!' : a / b;
```
agniya-i, line 57: Better to use `throw` statement to throw an exception 👍
```suggestion
      return b === 0 ? throw new Error('ERROR! You can`t divide by zero!') : a / b;
```

agniya-i (approved):

PR #1630, opened 2023-07-05
solution

tndeineka:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    reset(num) {
+      this.result = 0;
+
+      return this;
+    },
+    operate(callback, num) {
+      this.result = callback.call(this, num);
+
+      return this;
+    },
```
loralevitska, line 63: lets make work like this for more practice with `this`
```suggestion
    operate(callback, num) {
      callback(num);

      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(num) {
+      return this.result + num;
+    },
+    subtract(num) {
+      return this.result - num;
+    },
```
loralevitska, line 47: add 1 empty line between methods. Fix all cases
```suggestion

    add(num) {
      return this.result + num;
    },
    
    subtract(num) {
      return this.result - num;
    },
```

loralevitska (approved):
Almost perfect) Let's improve

PR #1640, opened 2023-07-10
create calculator function

Solovei-Andrii:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+      return calculator;
+    },
+
+    subtract(number) {
+      calculator.result -= number;
+
+      return calculator;
+    },
+
+    divide(number) {
```
volodymyr-soltys97, line 59: The 'divide' method could potentially cause a 'divide by zero' error. Please add a check to ensure the 'number' parameter is not zero before performing the division.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callBack, number) {
+      return callBack(number);
```
volodymyr-soltys97, line 44: The 'operate' method should update the 'result' property of the 'calculator' object. Right now it just returns the result of the callback without storing it. Consider updating this method to store the result of the callback.

volodymyr-soltys97 (changes requested):
Good job

volodymyr-soltys97 (approved):
Well done 🔥

PR #1641, opened 2023-07-10
solution

NataliMax:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
-  // write code here
+  const calculator = {
+
+    result: 0,
+
+    add(a) {
+      calculator.result += a;
+
+      return calculator.result;
+    },
```
VitaliyBondarenko1982, line 48:
```suggestion
   add(a) {
      this.result += a;

      return this;
    },
```

use this keyword instead calculator
same for other operations methods

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, x) {
+      // operation.bind(calculator, x);
+      operation(x);
```
VitaliyBondarenko1982, line 76:
```suggestion
      operation.call(this, x);
```

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #1642, opened 2023-07-10
01

Yuliia-Kyrychenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    calculator.result = 0;
+
+    return this;
+  };
+
+  calculator.operate = function(callback, number) {
+    calculator.result = callback(number);
+
+    return this;
+  };
```
loralevitska, line 60: Let's more practice with `this`: 
```suggestion
  operate(callback, value) {
      callback.call(this, value);

      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {};
```
loralevitska, line 40: You can immediately return the object instead of constantly duplicating `calculator.`
```suggestion
  return { ..... };
```

loralevitska (changes requested):
Almost done) Let's improve

loralevitska (approved):
Well done)

PR #1643, opened 2023-07-10
Solution

GeorgeSavostikov:


src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
+    },
+
+    add(number) {
+      const variable = calculator.result + number;
+
+      return variable;
+    },
+
+    subtract(number) {
+      const variable = calculator.result - number;
```
Anastasiia-Svintsova, line 63: Instead of creating a new variable, update the result directly.

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      const variable = calculator.result + number;
```
Anastasiia-Svintsova, line 57: Instead of creating a new variable, update the result directly.

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    method: '',
+    result: 0,
+    operate(method, number) {
+      calculator.method = method;
+      calculator.result = method(number);
```
Anastasiia-Svintsova, line 45: You are overwriting the result with only the current operation. You should instead apply the operation to the current result.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call?retiredLocale=uk

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    method: '',
+    result: 0,
+    operate(method, number) {
+      calculator.method = method;
```
Anastasiia-Svintsova, line 44: There is no need to store the method in the object.

Anastasiia-Svintsova (changes requested):

DarkMistyRoom (approved):

PR #1644, opened 2023-07-11
Solution

Illia-L:


src/makeCalculator.js
```diff
@@ -37,7 +37,27 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: (result, number) => result + number,
```
volodymyr-soltys97, line 42: The 'add', 'subtract', 'multiply', 'divide' methods should not receive the 'result' as a parameter. They should use 'this.result' instead.

volodymyr-soltys97 (changes requested):
Good job 👍 

PR #1645, opened 2023-07-13
task solution

AnVBondar:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+      return this;
+    },
+
+    operate(callback, value) {
+      return callback(value);
+    },
+
+    add(value) {
+      calculator.result += value;
```
volodymyr-soltys97, line 54: Instead of 'calculator.result', you should use 'this.result' to reference the current calculator object.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+      return calculator;
```
volodymyr-soltys97, line 56: Instead of returning 'calculator', you should return 'this' to keep the method chainable on the current calculator object.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      return callback(value);
```
volodymyr-soltys97, line 50: The 'operate' method should apply the operation on the calculator's result. Currently, it is not using the result for anything.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    multiply(value) {
+      this.result *= value;
+
+      return this;
+    },
+
+    divide(value) {
+      this.result /= value;
```
volodymyr-soltys97, line 74: The method `divide` does not handle division by zero. It can lead to Infinity as a result or even worse, crash the application. You should add a check to see if value is not zero before performing the division.

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      this[callback.name](value);
```
volodymyr-soltys97, line 50:
```suggestion
      callback.call(this, value)
```

volodymyr-soltys97 (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,50 @@
...
+  const calculator = {
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
```
anastasiiavorobiova, line 49: It would be safer to check if the callback is a function to avoid a potential TypeError. You could add a condition like: 'if (typeof callback !== 'function') { throw new Error('Callback must be a function'); }'.

anastasiiavorobiova (approved):
Nice work! &#128076;

PR #1646, opened 2023-07-13
js_advanced_calculator

KarolinaWiktoria:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (input) => calculator.result + input,
+
+    subtract: (input) => calculator.result - input,
```
darokrk, line 45: The 'subtract' method should modify the 'result' property instead of just returning the subtraction. You may want to refactor it to 'calculator.result -= input';

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+
+    multiply: (input) => calculator.result * input,
+
+    divide: (input) => calculator.result / input,
```
darokrk, line 49: The 'divide' method should modify the 'result' property instead of just returning the division. You may want to refactor it to 'calculator.result /= input'. Also, don't forget to handle the case when 'input' is zero to avoid division by zero error.

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (input) => calculator.result + input,
+
+    subtract: (input) => calculator.result - input,
+
+    multiply: (input) => calculator.result * input,
```
darokrk, line 47: The 'multiply' method should modify the 'result' property instead of just returning the multiplication. You may want to refactor it to 'calculator.result *= input';

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (input) => calculator.result + input,
```
darokrk, line 43: The 'add' method should modify the 'result' property instead of just returning the addition. You may want to refactor it to 'calculator.result += input';

darokrk (changes requested):
Looks good few fixes and we got this 😃

src/makeCalculator.js
KarolinaWiktoria: All right! Thanks. Already fixed :).

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add: function(input) {
+      this.result += input;
+    },
```
DorotaLeniecDev, line 45: You can use short version:
```suggestion
    add() {
      this.result += input;
    },
```

DorotaLeniecDev (approved):
Good job, check my suggestion, I am approving the task

PR #1648, opened 2023-07-15
Solution

Anmonnn:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+      result = 0;
+
+      return calculator;
+    },
+    subtract(value) {
+      result -= value;
+
+      return calculator;
+    },
+    divide(value) {
```
volodymyr-soltys97, line 63: You are not handling division by zero. This will lead to Infinity, which might not be the expected behavior. You should add a check to ensure the value is not zero before performing the division.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  const calculator = {
+    operate(callback, value) {
+      callback(value);
```
volodymyr-soltys97, line 44: Callback function in the operate method does not have any effect on the result. You should assign the return value of the callback to the result

volodymyr-soltys97 (changes requested):
Good job 👍 

src/makeCalculator.js
Anmonnn: If you fix it as you say, then the tests do not pass, or I misunderstood you.
image

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      return calculator;
+    },
+    subtract(value) {
+      result -= value;
+
+      return calculator;
+    },
+    divide(value) {
+      if (value !== 0) {
+        result /= value;
```
anastasiiavorobiova, line 65: In the divide function, you should throw an error when 'value' is 0 to avoid division by zero.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

Viktor-Kost (approved):

PR #1653, opened 2023-07-17
add task solution

andrew17752985:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add: function(num) {
+      this.result += num;
+
+      return this;
```
polosanya, line 45:
```suggestion
```
There is also no need to return this in operations. It's needed only in reset and operate methods

polosanya (changes requested):
Almost done!

Viktor-Kost (approved):

anastasiiavorobiova (approved):
Nice work

PR #1654, opened 2023-07-18
Solution 18072023

Dmytro-Budynskyi:


src/makeCalculator.js
```diff
@@ -37,7 +37,32 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      this.result += num;
```
marina-bidenko, line 44: It would be good to add data validation: to check if the input is a number before performing operations.

src/makeCalculator.js
```diff
@@ -37,7 +37,32 @@
...
+    divide(num) {
+      this.result /= num;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(operation, num) {
+      operation.call(this, num);
```
marina-bidenko, line 61: It's not clear how to use the `operate` function. The code should be self-explanatory or have comments explaining complex parts.

marina-bidenko (changes requested):
Fix the comments

Ivan-Shvorak (approved):
LGTM

PR #1655, opened 2023-07-18
advanced calculator v1

oltarasov:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const newObj = {
+    result: 0,
+
+    reset() {
+      newObj.result = 0;
```
volodymyr-soltys97, line 44: Instead of referring to 'newObj', use 'this' to refer to the current object.

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+      return this;
+    },
+
+    operate(callback, value) {
+      callback(value);
```
volodymyr-soltys97, line 50: In the 'operate' method, you should pass 'this' as a context to callback function. Correct way to call it would be 'callback.call(this, value)'.

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+      return this;
+    },
+
+    add(value) {
+      newObj.result = newObj.result + value;
```
volodymyr-soltys97, line 56: Instead of 'newObj.result = newObj.result + value;', you can simply use 'this.result += value;'.

volodymyr-soltys97 (changes requested):
Good job

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+    },
+
+    multiply(value) {
+      this.result *= value;
+
+      return this;
+    },
+
+    divide(value) {
+      if (value !== 0) {
```
volodymyr-soltys97, line 74: The divide method does not handle the case where the value is 0. It should throw an error or return a specific value instead of allowing an attempt to divide by zero.

volodymyr-soltys97 (changes requested):

volodymyr-soltys97 (approved):
Awesome 🔥

PR #1658, opened 2023-07-20
add calculator solution

PavloMykhalov:


src/makeCalculator.js
```diff
@@ -38,6 +38,32 @@
...
+    divide(a, b) {
+      return a / b;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(callback, number) {
+      this.result = callback(this.result, number);
```
anastasiiavorobiova, line 62: The 'operate' method should check that 'callback' is a function before calling it, to prevent potential errors.

src/makeCalculator.js
```diff
@@ -38,6 +38,32 @@
...
+      return a + b;
+    },
+    subtract(a, b) {
+      return a - b;
+    },
+    multiply(a, b) {
+      return a * b;
+    },
+    divide(a, b) {
+      return a / b;
```
anastasiiavorobiova, line 54: The 'divide' method should handle the case when 'b' is zero to prevent division by zero error.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

anastasiiavorobiova (approved):
Awesome!

PR #1663, opened 2023-07-24
solution

IvanOlecsandrovych:
Done)

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    },
+    subtract(value) {
+      return calculator.result - value;
+    },
+    multiply(value) {
+      return calculator.result * value;
+    },
+    divide(value) {
+      return calculator.result / value;
+    },
```
etojeDenys, line 63: let's use `this` instead of `calculator` here

etojeDenys (changes requested):

IvanFesenko (approved):
Well done

PR #1665, opened 2023-07-26
add task solution

4xmplme:


src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+    reset() {
+      if (this.validateOperation(this.reset)) {
+        this.result = 0;
+
+        return this;
+      }
+    },
+    operate(operation, value) {
+      if (this.validateOperation(operation) && this.validateValue(value)) {
+        return operation.call(this, value);
```
volodymyr-soltys97, line 79: You're not handling situations where the operation is not valid or the value is not valid. It would be better to throw an error or return a specific value (like null or false) in these cases or you just remove this condition

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+
+      return this;
+    },
+    multiply(value) {
+      this.result *= value;
+
+      return this;
+    },
+    divide(value) {
+      if (value > 0) {
```
volodymyr-soltys97, line 58: It's incorrect to perform division operation only when the value is greater than 0. Division by 0 is undefined but not illegal, and it's better to handle that case separately and notify the user about it instead of silently ignoring division by 0.

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Awesome!

PR #1666, opened 2023-07-26
Solution

Patryk-Filipiak:


src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
...
  */
 function makeCalculator() {
-  // write code here
+  return {
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    divide: (a, b) => a / b,
+    multiply: (a, b) => a * b,
+    result: 0,
+
```
AndrzejPec, line 46: Wow, didn't know that you can immediately return an object. Pretty clever!

AndrzejPec (approved):
I find it correct ⚡ ⚡ ⚡

src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
...
+    operate(operation, digit) {
```
JoyCoffeeAddict, line 47: `digit` is incorrect variable name, because you can operate on a number that consists of multiple digits, it would be better to call it `number` or `value`

JoyCoffeeAddict (changes requested):

choeqq (changes requested):
Please apply comment above and you're good to go 🚀

src/makeCalculator.js
Patryk-Filipiak: done :P

Patryk-Filipiak:
> Please apply comment above and you're good to go 🚀

done : )

JoyCoffeeAddict (approved):
Great job!

PR #1673, opened 2023-07-27
Develop

anpiet:
advancedCalculator Solution

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,6 @@
  * @return {object}
  */
 function makeCalculator() {
-
   return {
     result: 0,
 
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+
+  return {
+    result: 0,
+
+    add(num) {
+      this.result += num;
+
+      return this;
+    },
+
+    subtract(num) {
+      this.result -= num;
+
+      return this;
+    },
+
+    multiply(num) {
+      this.result *= num;
+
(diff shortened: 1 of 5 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      this.result += num;
+
+      return this;
+    },
```
AndrzejPec, line 47: I guess you could shorten it with arrow funciton ▶️

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, num) {
+      callback.call(this, num);
+
+      return this;
```
AndrzejPec, line 76: same as above - but I'm not sure.

AndrzejPec (changes requested):
little improvement suggested, good job 🎸 🥁

src/makeCalculator.js
anpiet: It's not possible to use arrow function here.

src/makeCalculator.js
anpiet: It's not possible to use arrow function here.

src/makeCalculator.js
JoyCoffeeAddict: @AndrzejPec That is not possible, because `this` keyword has different value within arrow functions.

JoyCoffeeAddict (approved):

choeqq (approved):

PR #1677, opened 2023-07-28
advanced_calculator-solution

olha-rypich:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      calculator.result *= number;
+    },
+
+    divide(number) {
+      calculator.result /= number;
+    },
+
+    reset() {
+      calculator.result = 0;
+
```
etojeDenys, line 61: let's use `this` instead of `calculator`

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Well done)

PR #1678, opened 2023-07-31
solution

Sancz0pansa:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    multiply(value) {
+      calculator.result *= value;
+
+      return this;
+    },
+
+    divide(value) {
+      calculator.result /= value;
```
choeqq, line 68: please make sure you can't divide by 0

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    result: 0,
+
+    operate(operation, value) {
+      operation(value);
+
+      return this;
+    },
+
+    add(value) {
+      calculator.result += value;
```
choeqq, line 50: you can use this to access object properties
```suggestion
      this.result += value;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+      return this;
```
choeqq, line 52: those methods don't have to return `this`, you will pass your method to `operate` method and then invoke your operation with value, so only operate method has to return `this`

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    subtract(value) {
+      calculator.result -= value;
```
choeqq, line 56:
```suggestion
      this.result -= value;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+      return this;
+    },
+
+    multiply(value) {
+      calculator.result *= value;
```
choeqq, line 62:
```suggestion
      this.result *= value;
```

src/makeCalculator.js
choeqq, line 68:
```suggestion
      this.result /= value;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    divide(value) {
+      calculator.result /= value;
+
+      return this;
+    },
+
+    reset() {
+      calculator.result = 0;
```
choeqq, line 74:
```suggestion
      this.result = 0;
```

choeqq (changes requested):
Almost there

choeqq (approved):
Looks good to me, nicely done ✅

PR #1679, opened 2023-07-31
solution

Dolnys:


src/makeCalculator.js
```diff
@@ -36,8 +36,52 @@
...
+    },
+
+    multiply(number) {
+      result *= number;
+
+      return this;
+    },
 
+    divide(number) {
+      if (number !== 0) {
```
choeqq, line 63: Nice check ✅

src/makeCalculator.js
```diff
@@ -36,8 +36,52 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
-}
+  let result = 0;
```
choeqq, line 41: That can also be an object property, then you’ll modowy this.result

src/makeCalculator.js
```diff
@@ -36,8 +36,52 @@
...
+        result /= number;
+      }
+
+      return this;
+    },
+
+    reset() {
+      result = 0;
+
+      return this;
```
choeqq, line 73: I think that only operate method needs to return this, you only pass those methods into operate

choeqq (changes requested):
Almost there 🚀

choeqq (approved):
Good job, all changes applied 🚀

PR #1681, opened 2023-07-31
init commit

domalewski:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+  calculator.divide = function(a, b) {
+    return a / b;
+  };
+
+  calculator.reset = function() {
+    this.result = 0;
+
+    return this;
+  };
```
choeqq, line 70: it would be better to add every method in object literal
```suggestion
 const calculator = {
  result = 0;

  operate(callback, param) {
    this.result = callback(this.result, param);

    return this;
  };

  add(a, b) {
    return a + b;
  };

  subtract(a, b) {
    return a - b;
  };

  multiply(a, b) {
    return a * b;
  };

  divide(a, b) {
    return a / b;
  };

  reset() {
    this.result = 0;

    return this;
  };
  
 }
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    return a - b;
+  };
+
+  calculator.multiply = function(a, b) {
+    return a * b;
+  };
+
+  calculator.divide = function(a, b) {
+    return a / b;
+  };
```
choeqq, line 64: you shouldn't be able to divide by 0, please take that into account 😄

choeqq (changes requested):
Almost there, small changes needed 🚀

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      if (b !== 0) {
+        return a / b;
+      } else {
+        return 'Error: Cannot divide by zero';
+      }
```
choeqq, line 67: you can just return outside of if statement - just return Error outside of function

choeqq (changes requested):

asakevych (approved):
Great job!!! 🚀 🔥 👏

PR #1685, opened 2023-08-01
add task solution

olya-girnyak:


maxim2310 (approved):
short and clear, nice work

PR #1687, opened 2023-08-02
task is done

vladkorobka:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    subtract(firstNum, secondNum) {
+      return firstNum - secondNum;
+    },
+
+    divide(firstNum, secondNum) {
+      if (secondNum !== 0) {
+        return firstNum / secondNum;
+      }
+
+      return this;
```
volodymyr-soltys97, line 56: Returning 'this' when division by zero is not appropriate. Should throw an error instead.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #1688, opened 2023-08-03
Solution

deandre25:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      this.result += num;
```
denys-danyliuk, line 44: Make it like a pure function.
For example
```
add(num1, num2) {
	return num1 + num2;
}

File: src/makeCalculator.js [hunk 2]
```diff
@@ -37,7 +37,45 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, num) {
+      callback.call(this, num);
```
denys-danyliuk, line 74: save your result here
```
this.result = callback

denys-danyliuk (changes requested):

xiimiik (approved):
Nice work!

PR #1690, opened 2023-08-03
solution

DenisChakhalian:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
```
Wesses, line 41: You can use const INITIAL_VALUE = 0;, because u use it twice.

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return result - number;
+    },
+
+    multiply(result, number) {
+      return result * number;
+    },
+
+    divide(result, number) {
+      return result / number;
+    },
```
Wesses, line 57: It's a bit weird to pass a `result `inside an object that already has a `result `field. But I don't know any other options so that you can not fix it.

Wesses (approved):
Well done.
Fix it or not, choose for yourself.

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
```
dffUqp, line 60: let's improve your code, I think we can avoid putting `this.result` directly in our callback.

dffUqp (changes requested):
Almost done:smiley:

dffUqp (approved):
LGTM🚀

PR #1691, opened 2023-08-03
add task solution

Wesses:
I do not know how to use `this `instead of `calculator`. I know when I am calling `calculator.operation()` inside of `this = calculator.operation`. I can not find out how to access the calculator object from a function, not using obvious `calculator.field_name`.

If u know, please leave the comment with solution, ty)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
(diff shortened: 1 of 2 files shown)


Wesses:
I fixed my problem, but still I don't mind additional explanations.

alexander-ignatow (approved):
LGTM 👍 

I don't quite get your comment - but if you still have questions make sure ut them in the chat or on QnA session.

PR #1692, opened 2023-08-03
js_advanced_calculator

MaksymShevch:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    operate(callback, value) {
```
volodymyr-soltys97, line 41: The 'operate' function is not checking if 'value' is a number. Consider adding validation to ensure 'value' is a number before performing the operation.

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      return this;
+    },
+    add(value, result) {
+      return result + value;
+    },
+    subtract(value, result) {
+      return result - value;
+    },
+    divide(value, result) {
+      return result / value;
```
volodymyr-soltys97, line 53: There's no check for division by zero. Ensure to handle this edge case.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      }
+    },
+    add(value, result) {
+      return result + value;
+    },
+    subtract(value, result) {
+      return result - value;
+    },
+    divide(value, result) {
+      if (value !== 0) {
```
volodymyr-soltys97, line 55: If 'value' is zero, the divide function will not return anything, leading to undefined behavior. It's better to throw an error or return a specific value.

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    operate(callback, value) {
+      if (!isNaN(value)) {
```
volodymyr-soltys97, line 42: There is no need to check whether the value is not a number. It's better to verify if the provided argument is indeed a number for better type safety.

volodymyr-soltys97 (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    operate(callback, value) {
+      if (Number.isInteger(value)) {
```
volodymyr-soltys97, line 42: Using operator `typeof` it will be better here

volodymyr-soltys97 (changes requested):
Still not fixed
image

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    operate(callback, value) {
```
volodymyr-soltys97, line 41: The 'operate' method should only take one argument. The second argument is unnecessary since 'this.result' is already accessible within the object.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+    subtract(value, result) {
+      return result - value;
+    },
+    divide(value, result) {
+      if (value !== 0) {
+        return result / value;
+      }
+
+      return 'error';
```
volodymyr-soltys97, line 59: Need to use `throw new Error` and you need to use more informative text for the error, the user should understand in detail what the error is

volodymyr-soltys97 (changes requested):

src/makeCalculator.js
MaksymShevch: i don't know, it doesn't work without value

MaksymShevch: The operate function cannot continue running if only one argument is passed to it, because it expects two arguments: callback and value. If only one argument is passed to it, then there will be an error inside the function, since an attempt to refer to a non-existent second argument will result in an error.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+    subtract(value, result) {
+      return result - value;
+    },
+    divide(value, result) {
+      if (value !== 0) {
+        return result / value;
+      }
+
+      return 'error: division by zero is not possible';
```
volodymyr-soltys97, line 59: Consider it will be better. You can read more here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw
```suggestion
      return throw new Error('error: division by zero is not possible');
```

volodymyr-soltys97 (approved):
Well done 🔥

src/makeCalculator.js
MaksymShevch: Thank you for the new information :)

MaksymShevch: image
but for some reason linter says problems...

PR #1695, opened 2023-08-05
Add task solution

davinnchii:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+      return this;
+    },
+    operate(callback, value) {
+      callback(value, this);
+
+      return this;
+    },
+    result: 0,
+  };
```
FrontKid, line 73: You can optimize your code using "bind" or "apply"
(https://learn.javascript.ru/bind)

FrontKid (approved):
Good job!!!

src/makeCalculator.js
alexander-ignatow: This is valid, but rather optional comment

alexander-ignatow (approved):
LGTM 👍

src/makeCalculator.js
davinnchii: > This is valid, but rather optional comment

I thought about implementing it, but was more focused on other tasks rather than trying to do it

PR #1696, opened 2023-08-05
add task solution

FrontKid:


maksym-mishchanchuk (approved):
GooooD!

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
+  const calculator = {
+    DEFAULT_VALUE: 0,
+    result: 0,
+
+    operate(
+      modifyResult = N => N ?? this.result,
+      n = this.DEFAULT_VALUE,
+    ) {
+      // add context to the callback function and call it
+      modifyResult.bind(this)(n);
```
dffUqp, line 49: We won't reuse this function. Why can't we just call it with our context. (let's use `call` instead of `bind`)

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
+
+      return this;
+    },
+
+    add(n) {
```
dffUqp, line 54: use more descriptive name for function argument

dffUqp (changes requested):
Almost done:smiley:

src/makeCalculator.js
FrontKid: I don't think it's really necessary, we always take a number as a parameter
So what's the best way to name a variable containing a random number

dffUqp: I think better to avoid using one-letter variable naming(It could be confusing for another developer, who reads your code)

So let's replace `n` with `number`, maybe a little bit, but it will improve readability of your code

dffUqp (approved):
✔️

PR #1697, opened 2023-08-05
initial commit

maksym-mishchanchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(action, number) {
+      action.call(this, number);
+
+      return this;
```
desevoker, line 70:
```suggestion
      return action.call(this, number);
```

Action already returns `this`.

desevoker (approved):
Goob job! 🚀

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(number) {
+      this.result += number;
+
+      return this;
```
dffUqp, line 45: what happens if we remove this line?

dffUqp (changes requested):
Almost done:smiley:

src/makeCalculator.js
maksym-mishchanchuk: ✕ 'operate' should support a chain call (1ms)
    ✕ 'operate' should correctly work in a long chain

i have 2 error after remove

dffUqp: I think you remove `return this` in `operate` method because I don't have these errors. 

image

dffUqp: 👀

dffUqp (changes requested):

src/makeCalculator.js
maksym-mishchanchuk: Снимок экрана 2023-08-09 в 09 20 50

maksym-mishchanchuk: if i delete s have problem with some task

is you remember, last time i have problem with some tasks test. I don*t know why

dffUqp (approved):

PR #1698, opened 2023-08-05
Solution

desevoker:


polinavafik (approved):
Well done! 🎀

src/makeCalculator.js
```diff
@@ -1,43 +1,40 @@
...
- *
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
```
dffUqp, line 11: let's improve your code, I think we can avoid putting `this.result` directly in our callback.

dffUqp (changes requested):
Almost done:smiley:

src/makeCalculator.js
desevoker: Ok, I changed call logic. Is it better now?

dffUqp (approved):

PR #1699, opened 2023-08-06
solution

polinavafik:
I had `return this` in each operation, but removed it since we would `return this ` anyway in `operate`


kaavka (commented):
Looks awesome!✨

Vannnkof (approved):

dffUqp (approved):
Perfect🚀

PR #1700, opened 2023-08-06
Add task solution

spojrzenie:


src/makeCalculator.js
```diff
@@ -36,8 +36,43 @@
...
+
+      return this;
+    },
+    add: function(number) {
+      calc.result += number;
+
+      return this;
+    },
+    subtract: function(number) {
+      calc.result -= number;
```
darokrk, line 54: Same here as above:

Instead of using `calc.result` use `this.result`. `this` keyword refers to the current object, providing flexibility if the object name is changed.

Please do that in all functions :)

src/makeCalculator.js
```diff
@@ -36,8 +36,43 @@
...
+      return this;
+    },
+    reset: function() {
+      calc.result = 0;
+
+      return this;
+    },
+  };
+
+  return calc;
```
darokrk, line 75: this line can be removed

src/makeCalculator.js
```diff
@@ -36,8 +36,43 @@
...
+      calc.result += number;
+
+      return this;
+    },
+    subtract: function(number) {
+      calc.result -= number;
+
+      return this;
+    },
+    multiply: function(number) {
```
darokrk, line 58: please add one space separator between function to make code cleaner to read :)

darokrk (changes requested):
Looks good, few improvements needed and we good to go 😃

darokrk (approved):
Nicely done 🔥  good job!

PR #1703, opened 2023-08-07
add task solution

vserhiichuk:


Orpheus29 (approved):
Yessss, this works, and the checklist is met.
Will be interesting to hear the mentors' opinion on the 'this / .call' use.

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(value) {
+      this.result += value;
+
+      return this;
```
dffUqp, line 45: what happend if we remove `return this` here?

dffUqp (changes requested):
Looks good👍 only one small fix needed

src/makeCalculator.js
vserhiichuk: Nothing😅, thanks, my bad

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      this.result /= value;
+    },
+    multiply(value) {
+      this.result *= value;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
```
dffUqp, line 58: To improve the readability of the code, we can add line break after each method.

dffUqp (approved):
🚀

PR #1704, opened 2023-08-07
add task solution

Vannnkof:


kaavka (commented):
looks awesome ✨

kaavka (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+
+    reset(integer) {
+      calculator.result = INITIAL_VALUE;
+
+      return this;
+    },
+
+    operate(action, integer) {
+      action(integer);
```
dffUqp, line 66: Recommendation:

dffUqp (changes requested):
Almost done:smiley:

alexander-ignatow (approved):
LGTM 👍

PR #1705, opened 2023-08-07
Solution

Orpheus29:


src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
AndriiKuz, line 40: Use `this` instead `calculator`

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
```
AndriiKuz, line 44: You change the result value in every methods. Will be better to do it only in operate method. In this method just return a + b

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(callback, number) {
+      return callback(number);
```
AndriiKuz, line 74: and return callback(number); replace : this.result = callback(this.result, number); return this;

AndriiKuz (changes requested):
Good job, but can be improved

src/makeCalculator.js
Orpheus29: I appreciate your help, Andrii. I didn't like the repetitions either, but couldn't get how to get rid of them.

dffUqp (approved):
LGTM🚀

PR #1708, opened 2023-08-07
Solution

neovendano:


KatOlista (approved):
good job!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+  const calculator = {
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, operand) {
```
dffUqp, line 49: Better name for function that we put inside `operate` method is `callback`


> A callback is a function passed as an argument to another function.

Learn more about callback  --> [link](https://www.w3schools.com/js/js_callback.asp) 

```suggestion
    operate(callback, operand) {
```

dffUqp (approved):
🔥

PR #1710, opened 2023-08-08
solution

Valentyn-Bilyk:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+      calculator.result -= number;
+    },
+
+    multiply(number) {
+      calculator.result *= number;
+    },
+
+    divide(number) {
+      calculator.result /= number;
+    },
```
Anna-Tkachenko, line 57: Each separate method should not know anything about `calculator` and `result`
Each method should only accept 2 numbers, complete some action and return the result

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      callback(number);
+
+      return this;
+    },
```
Anna-Tkachenko, line 69: This method should be the main one
It should call callback, get the result and change `result` property

Anna-Tkachenko: result = result of callback operation

FrontKid (approved):
Good job!

dffUqp (approved):
GJ✔️

PR #1711, opened 2023-08-08
Develop

AndriiKuz:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.7.12",
+    "@mate-academy/scripts": "^1.2.8",
     "eslint": "^5.16.0",
     "eslint-plugin-jest": "^22.4.1",
```
ivannaromanovych, line 21: You need to commit only those files that you changed manually
> git add src/<task_name>.js

Check this out [the guideline](https://github.com/mate-academy/js_task-guideline/blob/master/README.md)

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(prevNum, curNum) {
+      return prevNum - curNum;
+    },
+
+    multiply(prevNum, curNum) {
+      return prevNum * curNum;
+    },
+
+    divide(prevNum, curNum) {
+      return prevNum / curNum;
```
ivannaromanovych, line 56: I think you need to change those parameters names to something like firstOperand and secondOperand
And also I think you should use arrow functions here

src/makeCalculator.js
```diff
@@ -37,7 +37,28 @@
...
+    multiply: (firstOperand, secondOperand) => firstOperand * secondOperand,
+    divide: (firstOperand, secondOperand) => firstOperand / secondOperand,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, num) {
```
dffUqp, line 54: Better name for function that we put inside `operate` method is `callback`


> A callback is a function passed as an argument to another function.

Learn more about callback  --> [link](https://www.w3schools.com/js/js_callback.asp) 

```suggestion
    operate(callback, operand) {
```

dffUqp (approved):
🚀

PR #1715, opened 2023-08-08
add task solution

NazarVotkanych:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(num) {
```
AndriiKuz, line 43: You can use the arrow function in operation methods to write in one line.

AndriiKuz (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      this.result += num;
+
+      return this;
```
dffUqp, line 46: what happens if we remove line :46?

dffUqp (approved):
LGTM🚀

PR #1722, opened 2023-08-11
calculator

KatiaGamal:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+    reset: function() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate: function(callback, number) {
+      if (typeof callback === 'function') {
+        callback.call(this, number);
+      }
```
VitaliyBodnar, line 78: I don't see any sense for that condition, because callback should be every time with type of 'function'

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):
Good job!

PR #1724, opened 2023-08-13
Develop

Duster-1:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: number => calculator.result + number,
+
+    subtract: number => calculator.result - number,
+
+    divide: number => calculator.result / number,
```
volodymyr-soltys97, line 47: You need to add a check for 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: number => calculator.result + number,
+
+    subtract: number => calculator.result - number,
```
mykhalenych, line 45: check all methods

mykhalenych (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      if (b !== 0) {
```
volodymyr-soltys97, line 56: You are not returning anything when division by zero occurs. It's better to throw an error or return some specific value.

volodymyr-soltys97 (changes requested):

anastasiiavorobiova (approved):
Nice work!

PR #1728, opened 2023-08-14
solution

mykyta01:


src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
 function makeCalculator() {
   // write code here
+  const obj = {
+    result: 0,
+
+    add(x, y) {
+      return x + y;
+    },
+
+    subtract(x, y) {
```
mykhalenych, line 48: check all methods

mykhalenych (changes requested):

src/makeCalculator.js
```diff
@@ -38,6 +38,45 @@
  */
 function makeCalculator() {
   // write code here
+  const obj = {
+    result: 0,
+
+    add(a) {
+      this.result += a;
```
mykhalenych, line 45: All methods should return the object itself. You should add 'return this;' at the end of the method.

anastasiiavorobiova (approved):
Awesome work!

PR #1730, opened 2023-08-15
calculator

VasylKurtiak:


src/makeCalculator.js
```diff
@@ -36,8 +36,47 @@
...
+      calculator.result /= num;
+
+      return calculator;
+    },
+
+    reset: function() {
+      calculator.result = 0;
+
+      return calculator;
+    },
```
IvanFesenko, line 72: You should use `this` instead of calculator reference
```suggestion
    add: function(num) {
      this.result += num;

      return this;
    },

    subtract: function(num) {
      this.result -= num;

      return this;
    },

    multiply: function(num) {
      this.result *= num;

      return this;
    },

    divide: function(num) {
      this.result /= num;

      return this;
    },

    reset: function() {
      this.result = 0;

      return this;
    },
```

IvanFesenko (changes requested):
Good job, almost done. Refactor, you should use `this`

src/makeCalculator.js
VasylKurtiak: Ok

mykhalenych (approved):

PR #1735, opened 2023-08-16
Solution "Make advanced calculator"

tniklevych:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    result: 0,
+
+    operate(operation, number) {
+      operation(number);
+
+      return this;
+    },
+
+    add(x) {
+      calculator.result += x;
```
volodymyr-soltys97, line 50: Within the `add` method, `calculator.result` should be `this.result` to correctly refer to the result property of the current calculator instance.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(x) {
+      calculator.result -= x;
+    },
+
+    divide(x) {
+      calculator.result /= x;
+    },
+
+    multiply(x) {
+      calculator.result *= x;
```
volodymyr-soltys97, line 62: Within the `multiply` method, `calculator.result` should be `this.result` to correctly refer to the result property of the current calculator instance.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+      return this;
+    },
+
+    add(x) {
+      calculator.result += x;
+    },
+
+    subtract(x) {
+      calculator.result -= x;
```
volodymyr-soltys97, line 54: Within the `subtract` method, `calculator.result` should be `this.result` to correctly refer to the result property of the current calculator instance.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(operation, number) {
+      operation(number);
```
volodymyr-soltys97, line 44: The `operation` method is calling a function passed as an argument but it doesn't use the `this` keyword. It should be `this.operation(number)` to call the operation on the current calculator instance.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    add(x) {
+      calculator.result += x;
+    },
+
+    subtract(x) {
+      calculator.result -= x;
+    },
+
+    divide(x) {
+      calculator.result /= x;
```
volodymyr-soltys97, line 58: Within the `divide` method, `calculator.result` should be `this.result` to correctly refer to the result property of the current calculator instance. You also need to add a check for 0, since you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #1736, opened 2023-08-17
solution was added

Liliya-Kalinichenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(x) {
+      calculator.result -= x;
+    },
+
+    multiply(x) {
+      calculator.result *= x;
+    },
+
+    divide(x) {
+      calculator.result /= x;
```
Anastasiia-Svintsova, line 68: There should be a zero division check before divide operation. If x equals to 0, you should throw an error or return some message about impossible operation.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+      return this;
+    },
+
+    add(x) {
+      calculator.result += x;
+    },
+
+    subtract(x) {
+      calculator.result -= x;
```
Anastasiia-Svintsova, line 60: The calculator object should be referred to as this, not calculator. Replace calculator.result with this.result.

Anastasiia-Svintsova (changes requested):
Almost done!

PR #1739, opened 2023-08-20
advanced calculator

Q1W2E3R4T5Y6U7I8a:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    operate: function(callback, number) {
+      this.result = callback(this.result, number);
```
anastasiiavorobiova, line 44: Consider binding context to use this.result inside 'add' and other functions

anastasiiavorobiova: Also, create a check if the callback argument is valid and it is a function to prevent further errors

anastasiiavorobiova (changes requested):
Good work! Almost done

src/makeCalculator.js
Q1W2E3R4T5Y6U7I8a: 32ffa75571eb4adcda1ba963d24b5f7cf5ef3cdf
I changed the code, please check

Andriy-Shymkiv (approved):
🔥

PR #1743, opened 2023-08-22
add task solution

atrofymovych:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      this.result -= num;
+
+      return this;
+    },
+    multiply: function(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide: function(num) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

Viktor-Kost (approved):

PR #1745, opened 2023-08-23
calculator

shelyaa:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(value) {
+      this.result -= value;
+    },
+
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      this.result /= value;
```
volodymyr-soltys97, line 56: No check for division by zero. If 'value' is zero, the function will throw an error. Before dividing, check if the 'value' is zero and handle the error as you see fit.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      callback.call(this, value);
```
volodymyr-soltys97, line 66: No check for the type of 'value'. There should be a check to ensure that the 'value' is a number before using it in a mathematical operation to prevent potential errors.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -37,7 +37,63 @@
...
+    multiply(value) {
+      if (typeof value === 'number') {
+        this.result *= value;
+      } else {
+        return 'Value must be a number.';
+      }
+    },
+
+    divide(value) {
+      if (typeof value === 'number') {
```
anastasiiavorobiova, line 68: You have already checked that the value is a number inside the operate function

src/makeCalculator.js
```diff
@@ -37,7 +37,63 @@
...
+    subtract(value) {
+      if (typeof value === 'number') {
+        this.result -= value;
+      } else {
+        return 'Value must be a number.';
+      }
+    },
+
+    multiply(value) {
+      if (typeof value === 'number') {
```
anastasiiavorobiova, line 60: You have already checked that the value is a number inside the operate function

src/makeCalculator.js
```diff
@@ -37,7 +37,63 @@
...
+    add(value) {
+      if (typeof value === 'number') {
+        this.result += value;
+      } else {
+        return 'Value must be a number.';
+      }
+    },
+
+    subtract(value) {
+      if (typeof value === 'number') {
```
anastasiiavorobiova, line 52: You have already checked that the value is a number inside the operate function

src/makeCalculator.js
```diff
@@ -37,7 +37,63 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      if (typeof value === 'number') {
```
anastasiiavorobiova, line 44: You have already checked that the value is a number inside the operate function

src/makeCalculator.js
```diff
@@ -37,7 +37,63 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      if (typeof value === 'number') {
```
anastasiiavorobiova, line 86: Also could you please implement the check that callback is a function?

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #1747, opened 2023-08-24
solution

Patryk-Buczkowski:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(a) {
```
darokrk, line 43: Please change function parameter name to be more descriptive like, e.g `number` instead of `a`. Also think you should return this to keep current context `return this;`

```
add(number) {
   this.result += number;

   return this;
 },
```

```
subtract(number) {
   this.result -= number;
      
    return this;
 },
```

etc..

Please do that in all functions.

darokrk (changes requested):
Looks good 👍🏽  left few mentions for code improvements

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
```
OleziO, line 37: Remove all comments.

OleziO (changes requested):

Radoslaw-Czerniawski (approved):
Very good :)

PR #1754, opened 2023-08-30
solving the task

shikoryak:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(a, b) {
```
mykhalenych, line 42: The 'add', 'subtract', 'multiply' and 'divide' methods should not take two arguments. They should only update the 'result' with the given value.

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+      return a + b;
+    },
+    subtract(a, b) {
+      return a - b;
+    },
+    multiply(a, b) {
+      return a * b;
+    },
+    divide(a, b) {
+      return a / b;
```
mykhalenych, line 52: No handling for division by zero scenario. This should be handled to prevent the program from crashing.

mykhalenych (changes requested):

IvanFesenko (approved):
Well done

PR #1758, opened 2023-08-31
add task solution

kitpes:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(object, number) {
```
volodymyr-soltys97, line 43: The 'add', 'subtract', 'multiply' and 'divide' methods should not take 'object' as a parameter. They should operate on the 'result' property of the 'calculator' object directly.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    subtract(object, number) {
+      object.result -= number;
+    },
+
+    multiply(object, number) {
+      object.result *= number;
+    },
+
+    divide(object, number) {
```
volodymyr-soltys97, line 55: Before performing division, there should be a check if 'number' is not equal to 0 to avoid division by zero error.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
+
+      return this;
+    },
```
IvanFesenko, line 73: It will be better solution
```suggestion
    add(number) {
      this.result += number;
    },

    subtract(number) {
      this.result -= number;
    },

    multiply(number) {
     this.result *= number;
    },

    divide(number) {
      if (number === 0) {
        throw new Error('Division by 0');
      }

     this.result /= number;
    },

    reset() {
      this.result = 0;

      return this;
    },

    operate(callback, number) {
      callback.call(this, number);

      return this;
    },
```

IvanFesenko (approved):
Check my comment

PR #1759, opened 2023-08-31
add task solution

mishakoshilka:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+      return calculator;
+    },
+    multiply: (number) => {
+      result *= number;
+
+      return calculator;
+    },
+    divide: (number) => {
+      result /= number;
```
volodymyr-soltys97, line 59: There is no check for zero before division. This might lead to a 'Division by Zero' error.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
+
+      return calculator;
+    },
+    reset: () => {
+      result = 0;
+
+      return calculator;
+    },
+    operate: (callback, number) => {
+      callback(number);
```
volodymyr-soltys97, line 72: The 'operate' function does not affect the 'result' value. It should be implemented accordingly so as to operate on the result.

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
+      return calculator;
+    },
+    multiply: (number) => {
+      result *= number;
+
+      return calculator;
+    },
+    divide: (number) => {
+      if (result === 0) {
+        return 'You cannot divide by zero';
```
volodymyr-soltys97, line 60: Instead of returning a string when dividing by zero, it's better to throw an error to make it explicit that something went wrong.

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
+
+      return calculator;
+    },
+    multiply: (number) => {
+      result *= number;
+
+      return calculator;
+    },
+    divide: (number) => {
+      if (result === 0) {
```
volodymyr-soltys97, line 59: The condition to check if the result is zero isn't correct. It should check if the input number is zero before performing division operation, not the result variable.

src/makeCalculator.js
```diff
@@ -37,7 +37,48 @@
...
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  const calculator = {
+    add: (number) => {
+      result += number;
+
+      return calculator;
+    },
```
volodymyr-soltys97, line 47:
```suggestion
  const calculator = {
    let result = 0;
    
    add: (number) => {
      this.result += number;
    },
```

volodymyr-soltys97 (changes requested):

volodymyr-soltys97 (approved):

PR #1763, opened 2023-09-02
solution

Masha-Cactus:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    subtract(num) {
+      this.result -= num;
+    },
+
+    multiply(num) {
+      this.result *= num;
+    },
+
+    divide(num) {
+      this.result /= num;
```
Andriy-Shymkiv, line 56: what if `num` is 0? can we divide by zero?

Andriy-Shymkiv (changes requested):

OleziO (approved):

PR #1766, opened 2023-09-04
Solution

natamrshn:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: (num) => {
+      calculator.result += num;
```
anastasiiavorobiova, line 43:
```suggestion
      this.result += num;
      
      return this.result;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+
+      return calculator;
+    },
+    subtract: (num) => {
+      calculator.result -= num;
+
+      return calculator;
```
anastasiiavorobiova, line 50: The same here

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+    reset: () => {
+      calculator.result = 0;
+
+      return calculator;
+    },
+    operate: (operation, num) => {
+      operation(num);
```
anastasiiavorobiova, line 68: Consider binding context for the operation function

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      calculator.result /= num;
+
+      return calculator;
+    },
+    reset: () => {
+      calculator.result = 0;
+
+      return calculator;
+    },
+    operate: (operation, num) => {
```
anastasiiavorobiova, line 67: Consider implementing the check if the operation is a function and the num is a number

maxim2310 (approved):
Good work!!!

PR #1767, opened 2023-09-04
Solution

RomanPerepelkin:


src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+    add(number) {
+      return calculator.result + number;
+    },
+
+    subtract(number) {
+      return calculator.result - number;
+    },
+
+    divide(number) {
+      return calculator.result / number;
```
volodymyr-soltys97, line 53: You need to add a check for 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's to improve your code

src/makeCalculator.js
```diff
@@ -36,8 +36,43 @@
...
+
+    subtract(number) {
+      return calculator.result - number;
+    },
+
+    divide(number) {
+      if (number !== 0) {
+        return calculator.result / number;
+      }
+    },
```
volodymyr-soltys97, line 56: Now when there is 0, nothing will happen, you need to show a message that you cannot divide by 0

volodymyr-soltys97 (changes requested):

PR #1770, opened 2023-09-05
added tasck solution

Stako21:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(a) {
+      this.result -= a;
+    },
+
+    multiply(a) {
+      this.result *= a;
+    },
+
+    divide(a) {
+      this.result /= a;
```
volodymyr-soltys97, line 62: No handling for division by zero. This will cause an Infinity result or error. Add a condition to check if the divisor 'a' is zero before performing the operation.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(operation, a) {
+      operation.call(this, a);
+
```
volodymyr-soltys97, line 45: The operate function assumes 'a' will be a number. There is no error checking to ensure that 'a' is a number. Add a check to ensure that 'a' is a number before using it.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's to improve your code

nazarmatsevych (approved):

PR #1773, opened 2023-09-07
Advanced calculator. Tests passed

momos1703:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    subtract(result, value) {
+      return result - value;
+    },
+
+    multiply(result, value) {
+      return result * value;
+    },
+
+    divide(result, value) {
```
mykhalenych, line 55: You should add a check to prevent division by zero

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(result, value) {
```
mykhalenych, line 43: The add, subtract, multiply, and divide methods shouldn't take result as a parameter. They should operate on this.result

mykhalenych (changes requested):

src/makeCalculator.js
momos1703: I don't know why but if I try using this.result directly into methods, I can't pass tests...

Viktor-Kost: Try to use like this
```
add(value) {
  this.result += value;

  return this;
},
```

Viktor-Kost (changes requested):
Add comment
https://github.com/mate-academy/js_advanced_calculator/pull/1773#discussion_r1318639115

If you still have some troubles feel free to ask at fe_chat

src/makeCalculator.js
momos1703: It's not working too(

Viktor-Kost (approved):

PR #1774, opened 2023-09-08
solution

maximmorenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(a, b) {
+      return a + b;
+    },
```
OleziO, line 45:
```suggestion
    add(a) {
      this.result += a;
      
      return this.result;
    },
```

OleziO (changes requested):
There must be only one parameter in each method.

maxim2310 (approved):
Good work!!!

PR #1777, opened 2023-09-11
task solution

thesidorenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    add(num) {
+      this.result += num;
+    },
+
+    subtract(num) {
+      this.result -= num;
+    },
+
+    divide(num) {
```
anastasiiavorobiova, line 51: Consider throwing an error in case of division by 0

anastasiiavorobiova (changes requested):
Great work! Added tiny comment to fix

loralevitska (approved):
Well done)

PR #1778, opened 2023-09-11
add task solution

ghost:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    divide(x) {
+      calculator.result /= x;
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
```
etojeDenys, line 62: better to use `this` instead of `calculator`, to have the ability to reuse our object in the future

etojeDenys: don't forget to call/bind/apply the callback in operate

etojeDenys (changes requested):

src/makeCalculator.js
ololopos: done

ololopos:
fixed , please review

etojeDenys (approved):
awesome

PR #1779, opened 2023-09-12
task solution

TenetaJ:


src/makeCalculator.js
```diff
@@ -37,7 +37,57 @@
...
+    subtract,
+    multiply,
+    divide,
+    reset,
+    operate,
+    get result() {
+      return result;
+    },
+  };
+
```
IvanFesenko, line 89: Methods and result should be declared inside this object. Also ypou should work with `this` keyword

IvanFesenko (changes requested):
Please refactor using `this` keyword

src/makeCalculator.js
TenetaJ: done

lerastarynets (approved):
You're doing amazing, keep it up!

PR #1780, opened 2023-09-12
Solution

julwer1k:
image

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,26 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
+    },
+
+    add: (x, y) => x + y,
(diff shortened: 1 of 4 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
...
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
+    },
+
+    add: (x, y) => x + y,
+    subtract: (x, y) => x - y,
+    multiply: (x, y) => x * y,
+    divide: (x, y) => x / y,
```
maxim2310, line 52: what happens when y === 0?

maxim2310 (changes requested):

src/makeCalculator.js
julwer1k: .Yes, dividing by 0 will cause an error in the code, so you need to check for such cases. It seems that the rule correctly alerted in the presence of this operation and this number.

maxim2310 (approved):
Good work!!!

PR #1782, opened 2023-09-13
Solution

Wita-Shchurko:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    add: add,
+    subtract: subtract,
+    multiply: multiply,
+    divide: divide,
+    reset: reset,
+    result: 0,
+  };
+
+  function operate(operation, number = 0) {
+    operation(number);
```
mykhalenych, line 51: The 'this' keyword is not bound to any object inside the 'operate' function. Consider using bind, call or apply method to bind 'this' to the calculator object.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+
+    return this;
+  }
+
+  function add(number) {
+    calculator.result += number;
```
mykhalenych, line 57: Consider using 'this.result' instead of 'calculator.result' to make the code more maintainable and flexible.

mykhalenych (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+  function subtract(number) {
+    this.result -= number;
+  }
+
+  function multiply(number) {
+    this.result *= number;
+  }
+
+  function divide(number) {
+    this.result /= number;
```
mykhalenych, line 69: There is no check for division by zero. If number passed as argument is zero, it will lead to an error as division by zero is undefined.

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+  }
+
+  function reset() {
+    calculator.result = 0;
```
mykhalenych, line 73: add this here as well

mykhalenych (changes requested):

sTorba24 (approved):

PR #1786, opened 2023-09-15
Complete Task

Annbpiu:


src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      if (a === 0) {
```
maxim2310, line 56:
```suggestion
      if (b === 0) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+        return this;
```
maxim2310, line 57: and better to return error message

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

VitaliyBodnar (approved):
Good job!

PR #1788, opened 2023-09-16
add task solution

Vovchukkul:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      this.result *= number;
+
+      return this;
+    },
+
+    divide: function(number) {
+      if (number !== 0) {
+        this.result /= number;
+      } else {
+        return 'Division by zero is not allowed.';
```
etojeDenys, line 65: better to throw the error here

etojeDenys (approved):
lgtm

PR #1795, opened 2023-09-19
add solution

myuzkiv:


maxim2310 (changes requested):
need to handle case if in divide func enter 0

sTorba24 (approved):

PR #1798, opened 2023-09-21
`SOLUTION`

sviatikkondur:
Hello, check my work!

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,34 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  const calculatorObject = {
+  const calculator = {
     result: 0,
     add(value) {
-      calculatorObject.result += value;
+      calculator.result += value;
 
       return this;
     },
 
     subtract(value) {
-      calculatorObject.result -= value;
+      calculator.result -= value;
 
       return this;
     },
 
     multiply(value) {
-      calculatorObject.result *= value;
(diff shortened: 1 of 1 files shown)


Helen-arch (approved):
LGTM

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(action, number) {
+      action(number);
```
vasylmiskiv, line 73: You should bind your external function to the object for access to `this`

vasylmiskiv (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(action, number) {
+      action.bind(this)(number);
```
vasylmiskiv, line 73: Better to use `call` method for binding because it calls immediately and `number` should be as a second param. Take a look how [call ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call)method works

vasylmiskiv (changes requested):

vasylmiskiv (approved):
GJ

PR #1799, opened 2023-09-21
add calculator solution

Helen-arch:


o-drozzdyk (approved):
Good job)

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+
+      return this;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(number, currentResult) {
+      return currentResult + number;
```
vasylmiskiv, line 53: Actually you don't need to use `currentResult ` as a second param in your method. You should use access to your result property via `this`

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    },
+    subtract(number, currentResult) {
+      return currentResult - number;
```
vasylmiskiv, line 56: The same here according to the comment above

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    },
+    divide(number, currentResult) {
+      return currentResult / number;
```
vasylmiskiv, line 59: > The same here according to the comment above

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    },
+    multiply(number, currentResult) {
+      return currentResult * number;
```
vasylmiskiv, line 62: > The same here according to the comment above

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    operate(callback, number) {
+      this.result = callback(number, this.result);
```
vasylmiskiv, line 43: It would be much better to bind the `callback` function to the context via the `call` method, for example to access this (in this case this object)

vasylmiskiv (changes requested):

StasSokolov (approved):
Great job, i'm approving it!

PR #1801, opened 2023-09-22
add Make Calculator task solution

o-drozzdyk:


vlevik (approved):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(a, b) {
+      return a + b;
```
vasylmiskiv, line 56: You need to use `this` like you did it in reset method because now you just return the sum of two params but you need to change the result

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    subtract(a, b) {
+      return a - b;
```
vasylmiskiv, line 60: > You need to use this like you did it in reset method

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    multiply(a, b) {
+      return a * b;
```
vasylmiskiv, line 64: > You need to use this like you did it in reset method

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    divide(a, b) {
+      return a / b;
```
vasylmiskiv, line 68: > You need to use this like you did it in reset method

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      this.result = callback(this.result, number);
```
vasylmiskiv, line 44: callback should bind to the context via `call` method and should has two params. The first param is context (this) and the second param is arguments (`number` in this case). And the function will be have access to `this` (to your object in this case) and property and change it. Also you don't need to assign the result of callback to the property

vasylmiskiv (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      object.result -= number;
+    },
+
+    multiply(object, number) {
+      object.result *= number;
+    },
+
+    divide(object, number) {
+      object.result /= number;
+    },
```
StasSokolov, line 69: i think each method should return result, like `reset or operate`

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(object, number) {
+      object.result += number;
+    },
```
StasSokolov, line 57: let's use `this`

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

StasSokolov (approved):
Great job, i'm approving it!)

PR #1802, opened 2023-09-22
solution

vlevik:


src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
+    divide(num) {
+      calculator.result /= num;
+
+      return calculator;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
```
YevhenKharko, line 72: You can return 'this' instead of calculator, but it's OK!

YevhenKharko (approved):
Everything is OK but check comments:) Approved! Good job!

src/makeCalculator.js
polosanya: ➕

polosanya (approved):
Good job! I'm approving it, but pay attention to the comment)

PR #1805, opened 2023-09-22
add task solution

annabasenyuk:


mpohorenyi (approved):
Great Job!

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    reset() {
+      obj.result = 0;
+
+      return obj;
+    },
+
+    operate(calculator, num) {
+      calculator(num);
```
vasylmiskiv, line 74: You should bind the calculator callback to the current context via `call` method

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+
+    add(num) {
+      obj.result += num;
+
+      return obj.result;
```
vasylmiskiv, line 46: You should return just `this `  and to the same for the following operations

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+
+    add(num) {
+      obj.result += num;
```
vasylmiskiv, line 44: Use `this` instead of `obj`

vasylmiskiv (changes requested):

src/makeCalculator.js
annabasenyuk: Hi) thanks for the comments, I hope I understood everything correctly and corrected it)

PR #1809, opened 2023-09-23
Solution

GlibVK:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: (num) => {
+      calculator.result += num;
```
sTorba24, line 44: don't use `calculator` variable,
use `this`

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+
+      return calculator;
+    },
+
+    multiply: (num) => {
+      calculator.result *= num;
+
+      return calculator;
+    },
```
sTorba24, line 53:
```suggestion
   multiply(number, currentValue) {
      return currentValue * number;
   },
```

sTorba24: you can do smth like that and set value only inside operate function

sTorba24 (changes requested):

VitaliyBodnar (approved):

PR #1810, opened 2023-09-23
Calculator objects 1st try

ElenaGrasovskaya:


shchipina (approved):
Perfect ✨ Good job!

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, num) {
+      callback(this, num);
```
polosanya, line 50: You can use call method instead
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Function/call

PR #1811, opened 2023-09-23
/add task solution

Chumakv:


src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+      result -= value;
+
+      return this;
+    },
+    multiply: function(value) {
+      result *= value;
+
+      return this;
+    },
+    divide: function(value) {
```
nazarmatsevych, line 58:
```suggestion
    multiply: function(value) {
      result *= value;

      return this;
    },
    
    divide: function(value) {
```

Add blank line between methods

nazarmatsevych (approved):
Good work, left minor comment

PR #1812, opened 2023-09-24
solution

shchipina:


sintax1s (approved):
I think your solution is good.

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, value) {
+      callback(value);
```
vasylmiskiv, line 44: You should bind your callback function to the context via the `call` method for access to `this`, because now your callback function knows nothing about the object

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+      return this;
+    },
+
+    add(value) {
+      calculator.result += value;
```
vasylmiskiv, line 50: You should use `this` instead of `calculator` because you use the methods of the current object not different one. You should change it in the following methods as well

vasylmiskiv (changes requested):

PR #1820, opened 2023-09-27
Solution

RVDotsenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: value => {
+      calculator.result += value;
+
+      return calculator;
+    },
```
sTorba24, line 47: make it no-arrow function and use `this` instead of `calculator`

sTorba24 (changes requested):

src/makeCalculator.js
RVDotsenko: I wrote like that, but it doesn`t work
https://drive.google.com/file/d/1Vem52sPst6wgqpShVPphntuSvegl7nXS/view?usp=sharing

```suggestion
    add(value) {
      this.result += value;
    },
```

sTorba24: try so

sTorba24 (changes requested):

src/makeCalculator.js
RVDotsenko: The same: https://drive.google.com/file/d/1wwY0w8EI-ftEV9ls93YprIwYIjqA7b7v/view?usp=sharing

src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+    multiply(value) {
+      this.result *= value;
+    },
+    divide(value) {
+      this.result /= value;
+    },
+
+    operate(operation, operand) {
+      operation.call(this, operand);
```
TarasHoliuk, line 57: well done :+1:

TarasHoliuk (approved):

src/makeCalculator.js
RVDotsenko: Thanks)

PR #1823, opened 2023-09-28
Solution

ihor-pavlov:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+    divide: (number) => {
+      calculator.result /= number;
+
+      return calculator;
+    },
+    reset: () => {
+      calculator.result = 0;
+
+      return calculator;
```
Viktor-Kost, line 64:
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+    operate: (method, number) => {
+      return method(number);
```
Viktor-Kost, line 67:
```suggestion
      this.result = method(this.result, number);
      return this;
```

Viktor-Kost (changes requested):

src/makeCalculator.js
ihor-pavlov: tests don't pass after the proposed changes

src/makeCalculator.js
ihor-pavlov: tests don't pass after the proposed changes

src/makeCalculator.js
Viktor-Kost: ok, let it be like this. It's better to use `this` for calculating as well and reorganize object a little. But your variant is also ok

Viktor-Kost (approved):

PR #1828, opened 2023-10-03
Solution

YevgeniySavchuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      this.result /= number;
+    },
```
xiimiik, line 56: Methods must accept 2 numbers, the current value and the previous value and return just a value of operation. Pay attention to the use of the `operate` method in the examples

src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+
+    operate(callback, number) {
+      callback.call(this, number);
```
xiimiik, line 59: tip
```suggestion
      callback.call(this.result, number);
```

src/makeCalculator.js
YevgeniySavchuk: Could you please explain why  I need 2 numbers for these methods? I have a "result" as the first operand and a "number" as the second operand.

YevgeniySavchuk: In the task methods "add", "subtract", "multiply", and "divide" do not have to "return" anything.

src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+    subtract(number) {
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      this.result /= number;
```
anastasiiavorobiova, line 55: Consider implementing a check. It's not allowed to divide by 0.

anastasiiavorobiova (changes requested):
Great work! Added a tiny comment to fix

src/makeCalculator.js
anastasiiavorobiova: Not fixed

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    },
+
+    operate(callback, number) {
+      if (number !== 0) {
```
anastasiiavorobiova, line 59: The check should be implemented inside the division method. Other methods can accept zero as an argument

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+        callback.call(this, number);
+
+        return this;
+      }
+
+      return 'Infinity';
```
anastasiiavorobiova, line 65:
```suggestion

```

anastasiiavorobiova (changes requested):
Good progress!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      if (number === 0) {
+        this.result = Infinity;
```
anastasiiavorobiova, line 56:
```suggestion
        throw new Error('Division by 0 is forbidden)
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
YevgeniySavchuk: Thanks for the improvement!

anastasiiavorobiova (approved):
Awesome work!

PR #1834, opened 2023-10-05
add task solution

IShamkii:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+  return {
+    result: 0,
+
+    operate(callback, value) {
+      callback.call(this, value);
+
+      return this;
+    },
+
+    add(a = 0) {
```
Viktor-Kost, line 49: no needs to add a default value

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+      return this.result;
+    },
+
+    multiply(a = 0) {
+      this.result *= a;
+
+      return this.result;
+    },
+
+    divide(a = 0) {
```
Viktor-Kost, line 67: create a condition to avoid dividing by zero

Viktor-Kost (changes requested):
Great!
Just minor fixes needed

anastasiia-tilikina (approved):
Well done!

PR #1835, opened 2023-10-05
Solution

antonshtef:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const object = {
```
```suggestion
  return {
```
And never use names like `object` :)

Just FYI

TarasHoliuk (approved):

PR #1841, opened 2023-10-09
add task solution

Softjey:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+      }
+
+      return this;
+    },
+    multiply(...numbers) {
+      for (const num of numbers) {
+        this.result *= num;
+      }
+
+      return this;
```
VitaliyBodnar, line 66: Screenshot 2023-10-09 at 13 09 10

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+    result: 0,
+    operate(operation, ...numbers) {
+      operation.apply(this, numbers);
+
+      return this;
+    },
+    add(...numbers) {
+      for (const num of numbers) {
+        this.result += num;
+      }
```
VitaliyBodnar, line 50: method add should work only with one number, so doesn't make sense to user cycle `for` here
```suggestion
    add(num) {
      this.result += num;
```

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):
Good job!

PR #1843, opened 2023-10-10
Solution

yslyshchenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    },
+    subtract(num) {
+      this.result -= num;
+    },
+    multiply(num) {
+      this.result *= num;
+    },
+    divide(num) {
+      this.result /= num;
+    },
```
volodymyr-soltys97, line 58: You need to add a check for 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

VitaliyBodnar (approved):
Good job!

PR #1844, opened 2023-10-10
solution

ArtemSydun:


src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
+
+    add(x) {
+      calculator.result += x;
+
+      return this;
+    },
+    subtract(x) {
+      calculator.result -= x;
+
+      return this;
```
polosanya, line 56: There is no need to return this everywhere. Only operate and reset methods should do it
image

polosanya (approved):
Good job!

PR #1848, opened 2023-10-13
Homework: JS Advanced Calculator

ViktorBykh:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+      this.result *= value;
+
+      return this;
+    },
+
+    divide(value) {
+      this.result /= value;
+
+      return this;
+    },
```
sarakhmen, line 65: Did you anticipate the possibility of division by zero? Let's add the corresponding validation

sarakhmen (changes requested):
Let's slightly improve your solution :)

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+
+    divide(value) {
+      if (value <= 0) {
+        return this;
+      } else {
+        this.result /= value;
+      }
+
+      return this;
```
sarakhmen, line 68: please try to refactor this piece of code and reduce number of return statements. And also pay attention that we should handle only division by zero, and negative numbers should be accepted

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #1849, opened 2023-10-14
first solution

Clover247:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      return callback(number);
+    },
+
+    add(b) {
+      calculator.result += b;
```
vasylmiskiv, line 48: The method inside the object so you need to use `this` instead of calculator
And do it the same to the following methods because `result` is internal property not in a different object

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      return callback(number);
```
vasylmiskiv, line 44: You need to bind your callback function to your object via call or apply method

vasylmiskiv (changes requested):

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #1855, opened 2023-10-16
solution

Radymyr:
why not work like this?
```
let calc = {
  result: 0,

  operate(callback, value) {
    this.result = callback(value);
    return this;
  },

  add(value) {
    this.result += value;
  },
};

calc.operate(calc.add, 10);
console.log(calc.result); 
```
```

this.result += value;
      ^

```

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, value) {
+      this.result = callback(this.result, value);
+
+      return this;
+    },
+
+    add(value1, value2) {
+      return value1 + value2;
+    },
+
+    subtract(value1, value2) {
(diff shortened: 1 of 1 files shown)


TarasHoliuk:
> why not work like this?

because when you do this:
`calc.operate(calc.add, 10);`

you pass the callback (calc.add) just as a function, w/o connection to the object. In other words, it looses `this`

to fix it you may use methods to define `this` inside a function when you call the passed callback inside the `operate` method (call/bind/apply)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
```suggestion
  return {
```
FYI: you could immediately return it

TarasHoliuk (approved):

PR #1856, opened 2023-10-16
Calculator

stunnerboyone:


src/makeCalculator.js
```diff
@@ -38,6 +38,49 @@
...
+      calculator.result -= number;
+
+      return calculator;
+    },
+
+    multiply(number) {
+      calculator.result *= number;
+
+      return calculator;
+    },
```
etojeDenys, line 70: let's use `this` here instead of `calculator`

src/makeCalculator.js
```diff
@@ -38,6 +38,49 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+
+    operate(method, number) {
+      return method(number);
```
etojeDenys, line 45: do not forget to bind this method to `this`

etojeDenys (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1858, opened 2023-10-17
Solution

FIREDAN99:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+      this.result *= num;
+
+      return calculator;
+    },
+
+    divide(num) {
+      this.result /= num;
+
+      return calculator;
+    },
```
VitaliyBodnar, line 65: don't need to return calculator object in these methods
please check again an example

Screenshot 2023-10-17 at 14 00 14

```suggestion
    add(num) {
      this.result += num;
    },

    subtract(num) {
      this.result -= num;
    },

    multiply(num) {
      this.result *= num;
    },

    divide(num) {
      this.result /= num;
    },
```

VitaliyBodnar (changes requested):

maxim2310 (approved):
Good work!!!

PR #1859, opened 2023-10-17
js_advanced_calculator

MariannaBo:


src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  const calculator = {
+
+    add: function(number) {
```
sTorba24, line 44: it should be object field 
```suggestion
  

  const calculator = {
    result: 0,
    
    add: function(number) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+      result += number;
+
+      return calculator;
```
```suggestion
      this.result += number;
```

sTorba24: you don't need to return something from operations

src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    reset: function() {
+      result = 0;
+
+      return calculator;
+    },
+    operate: function(operation, number) {
+      operation(number);
+
+      return calculator;
+    },
```
sTorba24, line 73:
```suggestion
    reset: function() {
      this.result = 0;

      return this;
    },
    operate: function(operation, number) {
      operation(number);

      return this;
    },
```

sTorba24 (changes requested):

anastasiiavorobiova (approved):
Awesome work!

PR #1866, opened 2023-10-18
Solution

VictoriaPukalo:


maxim2310 (changes requested):
Good work, but need to add exception for divide func when n = 0

etojeDenys (approved):
lgtm

PR #1867, opened 2023-10-18
Advanced calculator

kate0427:


src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+      calculator.result = callback(calculator.result, number);
+
+      return calculator;
+    },
+
+    reset: () => {
+      calculator.result = 0;
+
+      return calculator;
+    },
```
etojeDenys, line 66: better to use `this` instead of the `calculator`

etojeDenys (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,36 @@
...
+    subtract(result, number) {
+      return result - number;
+    },
+    multiply(result, number) {
+      return result * number;
+    },
+    divide(result, number) {
+      return result / number;
+    },
```
volodymyr-soltys97, line 54: You need to add a check for 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #1868, opened 2023-10-19
Solution

andreysakhno:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      this.result /= number;
+    },
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

VitaliyBodnar (approved):

PR #1872, opened 2023-10-22
calculator

Romanov7788:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(num) {
+      calculator.result += num;
+
+      return calculator;
```
anastasiiavorobiova, line 45:
```suggestion

```
There is no need to return the calculator from the method add/subtract / etc. We don't need to call them in a chain

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+    subtract(num) {
+      calculator.result -= num;
+
+      return calculator;
```
anastasiiavorobiova, line 50:
```suggestion

```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+    multiply(num) {
+      calculator.result *= num;
+
+      return calculator;
```
anastasiiavorobiova, line 55:
```suggestion

```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    },
+    divide(num) {
+      calculator.result /= num;
+
+      return calculator;
```
anastasiiavorobiova, line 60:
```suggestion

```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      calculator.result -= num;
+
+      return calculator;
+    },
+    multiply(num) {
+      calculator.result *= num;
+
+      return calculator;
+    },
+    divide(num) {
```
anastasiiavorobiova, line 57: Consider implementing a check if the num is not 0, to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

maxim2310 (approved):
Good work!!!

PR #1873, opened 2023-10-24
Solution

legogoUA:


src/makeCalculator.js
```diff
@@ -37,7 +37,46 @@
...
+    },
+    divide: function(num) {
+      result /= num;
+
+      return calculator;
+    },
+    reset: function() {
+      result = 0;
+
+      return calculator;
```
VitaliyBodnar, line 66: - it's better to return `this` instead of `calculator` as it will make a code more scalable 
- only reset and operate should return this, other method should't. Check again the example.

Screenshot 2023-10-24 at 17 34 30

VitaliyBodnar (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    multiply(num) {
+      calculator.result *= num;
+    },
+
+    subtract(num) {
+      calculator.result -= num;
+    },
+
+    divide(num) {
```
maxim2310, line 53: need to add check if num === 0

maxim2310 (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,53 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(number) {
+      if (number === 0) {
+        return calculator;
+      }
```
etojeDenys, line 45: actually, it's redundant +0, it doesn't affect the calculator at all

src/makeCalculator.js
```diff
@@ -37,7 +37,53 @@
...
+      calculator.result += number;
+
+      return calculator;
+    },
+    subtract(number) {
+      if (number === 0) {
+        return calculator;
+      }
```
etojeDenys, line 53: same here

src/makeCalculator.js
```diff
@@ -37,7 +37,53 @@
...
+      calculator.result -= number;
+
+      return calculator;
+    },
+    multiply(number) {
+      if (number === 0) {
+        calculator.result = 0;
+
+        return calculator;
```
etojeDenys, line 62: same here

src/makeCalculator.js
etojeDenys: still not fixed comments

etojeDenys (changes requested):

PR #1880, opened 2023-10-28
completed makeCalculator function

MrHunterBoi:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(instance, num) {
+      instance.result += num;
+    },
+    subtract(instance, num) {
```
polosanya, line 45:
```suggestion
    result: 0,
    
    add(instance, num) {
      instance.result += num;
    },
    
    subtract(instance, num) {
```
Don't forget to add newlines for better readability

src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    divide(instance, num) {
+      instance.result /= num;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(calcAction, num) {
+      calcAction(this, num);
```
polosanya, line 60: You can use call method instead
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

polosanya (changes requested):
Almost perfect!

MrHunterBoi:
Right, changed code accordingly xd

yevtok (approved):
Good job!

ArtemTeslenko (approved):
Good job 👍

PR #1887, opened 2023-10-31
Solution

MykhailoLoniak:


src/makeCalculator.js
```diff
@@ -36,8 +36,40 @@
...
+
+    subtract: function(currentResult, number) {
+      return currentResult - number;
+    },
+
+    multiply: function(currentResult, number) {
+      return currentResult * number;
+    },
+
+    divide: function(currentResult, number) {
```
maxim2310, line 67: need to add check if number === 0

maxim2310 (changes requested):
Good work, but lets make some improvement

maxim2310 (approved):
Good work!!!

PR #1892, opened 2023-11-01
add task solution

heorhiig:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add: function(num) {
+      this.result += num;
+
+      return this;
+    },
```
etojeDenys, line 47:
```suggestion
    add(num) {
      this.result += num;

      return this;
    },
```
we can simplify it

etojeDenys (approved):
lgtm

PR #1893, opened 2023-11-01
Solution

tania-kuzmenko:


src/makeCalculator.js
```diff
@@ -36,8 +36,50 @@
...
+      obj.result -= arg;
+
+      return obj.result;
+    },
+
+    multiply(arg) {
+      obj.result *= arg;
+
+      return obj.result;
+    },
```
etojeDenys, line 73: better to use `this` instead of the `obj`

src/makeCalculator.js
```diff
@@ -36,8 +36,50 @@
...
  */
+
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+    methods: ['add', 'subtract', 'multiply', 'divide', 'reset', 'operate'],
+
+    operate(method, arg) {
+      method(arg);
```
etojeDenys, line 46: and do not forget to bind this method

etojeDenys (changes requested):

src/makeCalculator.js
```diff
@@ -36,8 +36,50 @@
  *
  * @return {object}
  */
+
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+    methods: ['add', 'subtract', 'multiply', 'divide', 'reset', 'operate'],
```
VitaliyBodnar, line 43: **methods** field is not used anywhere, so it's redundant

src/makeCalculator.js
```diff
@@ -36,8 +36,50 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(arg) {
+      this.result += arg;
+
+      return this.result;
```
VitaliyBodnar, line 60: return this should be only for operate and reset methods, please check again the task description
Screenshot 2023-11-06 at 18 54 24
```suggestion
```

VitaliyBodnar (changes requested):

etojeDenys (approved):
good job

PR #1896, opened 2023-11-02
Solution

Patrykmclaren14:


JoyCoffeeAddict (changes requested):
Your code is almost perfect, but there is a funny catch, try to run this code
```
const c1 = makeCalculator()
const c2 = makeCalculator()
c1.operate(c2.add, 10)
console.log(c1.result);
```
So in normal circumstances your code would be good enough, but let's try to push ourselves to solve complex problems! 
You can try to work with `call bind and apply`, also think about `this` keyword

WiktoriaBielec (approved):
Great code 🔢

mvjl000 (approved):
Good

PR #1903, opened 2023-11-03
add task solution

OstinM:


src/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Document</title>
+</head>
+<body>
+ass
```
MrHunterBoi, line 9: Boobs better

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+    multiply(x, y) {
+      return x * y;
+    },
+
+    divide(x, y) {
+      if (y !== 0) {
+        return x / y;
+      } else {
+        throw new Error('Cannot divide by 0');
+      }
```
MrHunterBoi, line 67: You can simplify this by removing `else `statement

MrHunterBoi (approved):
Nice

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, value) {
+      calculator.result = callback(calculator.result, value);
+
+      return calculator;
```
ArtemTeslenko, line 47: It is better to use `this` in such cases, to get right context here you can use `call`

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    getResult() {
+      return calculator.result;
+    },
```
ArtemTeslenko, line 78: the same here, use `this`

ArtemTeslenko (changes requested):
Good job! Remove extra code, use `this`)

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(call, value) {
+      this.result = call(this.result, value);
+
+      return calculator;
```
artemforsoff, line 47: you can try to use `this` here

src/makeCalculator.js
```diff
@@ -36,8 +36,49 @@
...
+        return x / y;
+      }
+
+      throw new Error('Cannot divide by 0');
+    },
+
+    reset() {
+      this.result = 0;
+
+      return calculator;
```
artemforsoff, line 73: you can try to use `this` here also

artemforsoff (changes requested):
Great, but there are a couple places to improve and it would be perfect 🚀

BudnikOleksii (approved):
Good job 👍

PR #1909, opened 2023-11-06
add solution

MKLNKs:


lerastarynets (approved):
Well done! Short and clear, I like it)

PR #1911, opened 2023-11-06
Add task solution

kate-prakofyeva:


JoyCoffeeAddict (approved):
Great job, especially using `call` correctly

PR #1913, opened 2023-11-09
created makeCalculator function

mixelio:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
+    },
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1914, opened 2023-11-09
Develop

Anton-Kuchmasov:


ksyvvkaa (approved):
good job!

artemforsoff (changes requested):
Good job, try declaring calculator functions as object methods to work with `this`.
For example:
```
const user = {
    name: 'Anton',
    getName() {
        return this.name;
    },
};
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      callback(number);
+
+      return calculator;
```
artemforsoff, line 46: you can return `this` here
this would be better because you don't need to remember the name of the variable `calculator` in order to access it
also, if you have to rename it, you will have to do it in one place

artemforsoff (approved):
overall great job 🤟🏻

PR #1917, opened 2023-11-11
Solution

RostyslavKuts:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      calculator.result += number;
```
VitaliyBondarenko1982, line 56:
```suggestion
      this.result += number;
```
use this.restult in all operations methods

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    operate(callback, number) {
+      callback(number);
```
VitaliyBondarenko1982, line 44:
```suggestion
      callback.call(this, number);
```
after use this in all operations methods when method will be pass as argument to operate method - it will lose this.
for fix it we can use call method.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments.

src/makeCalculator.js
RostyslavKuts: Oh, sorry.
I never use callback.call()
 I can't understand how it works

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    add(number) {
+      this.result += number;
+    },
+
+    subtract(number) {
+      this.result -= number;
+    },
+
+    divide(number) {
```
anastasiiavorobiova, line 63: Try to implement a check to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

etojeDenys (approved):
good job

PR #1920, opened 2023-11-13
add soulution

Kozubowicz:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    },
+    subtract(x) {
+      calculator.result -= x;
+    },
+    multiply(x) {
+      calculator.result *= x;
+    },
+    divide(x) {
+      calculator.result /= x;
+    },
```
volodymyr-soltys97, line 53: It is necessary to provide for a check on 0, since it is not possible to divide by 0

volodymyr-soltys97 (changes requested):
Great work 🔥 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #1923, opened 2023-11-14
js_advanced_calculator by Dimoshua

Dimoshua:
https://github.com/Dimoshua/js_advanced_calculator

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -40,26 +40,41 @@
 function makeCalculator() {
   const calculator = {
     result: 0,
-    add(a, b) {
-      return a + b;
+
+    add: function(number) {
+      calculator.result += number;
+
+      return calculator;
     },
-    subtract(a, b) {
-      return a - b;
+
+    subtract: function(number) {
+      calculator.result -= number;
+
+      return calculator;
     },
-    multiply(a, b) {
-      return a * b;
+
+    multiply: function(number) {
+      calculator.result *= number;
+
+      return calculator;
     },
-    divide(a, b) {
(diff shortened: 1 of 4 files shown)


src/makeCalculator.js
```diff
@@ -36,8 +36,47 @@
...
+
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add: function(number) {
+      calculator.result += number;
+
+      return calculator;
```
maxim2310, line 47: better to use
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -36,8 +36,47 @@
...
+    },
+
+    multiply: function(number) {
+      calculator.result *= number;
+
+      return calculator;
+    },
+
+    divide: function(number) {
+      calculator.result /= number;
```
maxim2310, line 63: need to check if number === 0

maxim2310 (changes requested):
Good work, but lets make some improvement

src/makeCalculator.js
Dimoshua: if replaced then this error pops up:
● sum › 'operate' should support a chain call 

NOT better to use

src/makeCalculator.js
Dimoshua: divide: function(number) {
      if (number !== 0) {
        calculator.result /= number;
      }
  
      return calculator;
    },

maxim2310 (approved):
Good work!!!

PR #1926, opened 2023-11-15
Solution

deonara:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+      return calculator;
+    },
+    reset() {
+      result = 0;
+
+      return calculator;
+    },
+    get result() {
+      return result;
+    },
```
VitaliyBodnar, line 70: this getter is redundant, result is supposed  be a property of calculator object

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  const calculator = {
+    add(res = 0, num) {
+      result += num;
+
+      return calculator;
```
VitaliyBodnar, line 46: I suggest you to return this instead of calculator as it will make your code more scalable  
```suggestion
    add(res = 0, num) {
      result += num;

      return this;
```

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1929, opened 2023-11-16
Solution

InnaLysenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number = 0) {
+      callback(number);
+
+      return this;
+    },
+    add(number) {
+      calculator.result += number;
```
etojeDenys, line 48: let's use `this` instead of the calculator, it will improve reusability

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, number = 0) {
+      callback(number);
```
etojeDenys, line 43: do not forget to bind `this` to callback

etojeDenys (changes requested):

src/makeCalculator.js
InnaLysenko: Thanks. Done.

etojeDenys (approved):
good job

PR #1930, opened 2023-11-17
Develop

ptbit:
add solution

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -1,43 +1,39 @@
 'use strict';
 
-/**
- * Another calculator. Now the task is more difficult.
- * Create a `makeCalculator` function that returns an object that
- * has the following fields:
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -1,43 +1,31 @@
...
+    },
+    subtract(result, value) {
+      return result - value;
+    },
+    multiply(result, value) {
+      return result * value;
+    },
+    divide(result, value) {
+      return result / value;
+    },
```
volodymyr-soltys97, line 22: You need to add a check for 0, since we cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

OleziO (approved):

PR #1934, opened 2023-11-17
add task solution

maks2809:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+      return this;
+    },
+    multiply: function(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide: function(num) {
+      this.result /= num;
```
anastasiiavorobiova, line 58: Consider implementing a check if the number is not a 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    multiply: function(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide: function(num) {
+      if (num !== 0) {
+        this.result /= num;
+      } else {
+        return 'Error: Division by zero is not allowed';
```
VitaliyBodnar, line 61: it's better to return an error object for making errors 
```suggestion
        return trow new Error( 'Error: Division by zero is not allowed');
```

VitaliyBodnar (changes requested):

PR #1939, opened 2023-11-21
add task solution

reznik-denis:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    },
+    add(number, operator) {
+      return operator + number;
+    },
+    subtract(number, operator) {
+      return operator - number;
+    },
+    divide(number, operator) {
+      return operator / number;
+    },
```
Viktor-Kost, line 60: avoid dividing by zero. You can throw an error for this case

Viktor-Kost (changes requested):

src/makeCalculator.js
reznik-denis: I add error but I don't understand why it is there. there is no limits in task about that. And I am not sure that I do it correct.

VitaliyBodnar (approved):
Good job!

src/makeCalculator.js
VitaliyBodnar: You did it correct.

PR #1943, opened 2023-11-24
Solution

verandetta:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    subtract(num, pastNum) {
+      return pastNum - num;
+    },
+
+    multiply(num, pastNum) {
+      return pastNum * num;
+    },
+
+    divide(num, pastNum) {
```
anastasiiavorobiova, line 67: Consider adding a check: the number shouldn't be zero

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #1944, opened 2023-11-24
Solution

wsadocha:


src/makeCalculator.js
```diff
@@ -38,6 +38,33 @@
...
+    },
+    subtract(a) {
+      calculator.result -= a;
+    },
+    multiply(a) {
+      calculator.result *= a;
+    },
+    divide(a) {
+      calculator.result /= a;
+    },
```
volodymyr-soltys97, line 64: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #1945, opened 2023-11-25
Solution

KaterynaEV:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    multiply: function(num) {
+      this.result *= num;
+
+      return this;
+    },
+
+    divide: function(num) {
+      this.result /= num;
```
VitaliyBodnar, line 62: number can be 0, so please add a additional condition for this case
return some error message when num is 0

VitaliyBodnar (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1947, opened 2023-11-26
calaulator

oleksandr-rudiuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+
+    operate(callback, y) {
+      callback.call(this, this.result, y);
+
+      return this;
+    },
+    add(x, y) {
+      n = (x + y);
+      calculator.result = n;
+    },
```
VitaliyBodnar, line 52: Can be less, check other places 
```suggestion
    add(x, y) {
      this.result = x + y;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let n = 0;
```
VitaliyBodnar, line 40: This variable is redundant

VitaliyBodnar (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+
+  const calculator = {
+    operate(callback, y) {
+      callback.call(this, this.result, y);
```
anastasiia-tilikina, line 44: you don't need to pass this.result as a param

src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+
+      return this;
+    },
+    add(x, y) {
```
anastasiia-tilikina, line 48: you can pass only one param to your methods and make it work like this:

```javascript
this.result += x
``` 

fix everywhere

src/makeCalculator.js
anastasiia-tilikina: not fixed.

anastasiia-tilikina (changes requested):
just a couple more fixes

Viktor-Kost (changes requested):
.

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    },
+    add(x) {
+      this.result += x;
+    },
+    subtract(x) {
+      this.result -= x;
+    },
+    divide(x) {
+      this.result /= x;
+    },
```
Viktor-Kost, line 54:
```suggestion
    divide(x) {
      if (x === 0) {
        throw new Error("Division by zero is not allowed");
      }
      this.result /= x;
    },
```

Viktor-Kost (approved):
Nice!
Just apply recommendation I left for you

PR #1948, opened 2023-11-26
solution

stanislav-bzhezinskyi:


stanislav-bzhezinskyi:
Hi there! I need a hint, because my code is already working in my VScode, but I can't pass the tests.
And I also used 'this' keyword in my solution... for example 'return this' against 'return calculator'...
Anyway I need help. Thank you

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+
+const calculator = makeCalculator();
+
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(value) {
+      calculator.result += value;
```
etojeDenys, line 47: try to use `this` instead of the calculator

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+    },
+
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+
+    operate(action, value) {
+      action(value);
```
etojeDenys, line 69: and don't forget to bind the action

etojeDenys (changes requested):

etojeDenys:
> Hi there! I need a hint, because my code is already working in my VScode, but I can't pass the tests. And I also used 'this' keyword in my solution... for example 'return this' against 'return calculator'... Anyway I need help. Thank you

feel free to ask for some help in the chat

PR #1951, opened 2023-11-28
makeCalc

baraban2003:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    },
+
+    multiply: (num) => {
+      calculator.result *= num;
+
+      return calculator;
+    },
+
+    divide: (num) => {
+      calculator.result /= num;
```
anastasiiavorobiova, line 62: The division by 0 is forbidden, consider implementing a check here

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
```diff
@@ -37,7 +37,51 @@
...
+    },
+
+    divide: (num) => {
+      if (calculator.result === 0) {
+        calculator.result = 0;
+
+        return calculator;
+      }
+
+      calculator.result /= num;
```
anastasiiavorobiova, line 68: Not fixed. The division by 0 is still possible here

anastasiiavorobiova (changes requested):
Great work! Almost done

src/makeCalculator.js
baraban2003: but return 0. What can I do to fix it?

anastasiiavorobiova (approved):
Awesome!

PR #1955, opened 2023-11-29
added task solution

AlNeon:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(x, y) {
+      return x - y;
+    },
+
+    multiply(x, y) {
+      return x * y;
+    },
+
+    divide(x, y) {
+      if (y !== 0) {
```
l4st1m0za, line 56: Function 'divide' should return some value or throw an error when dividing by zero. It's currently not returning anything which could cause undefined behavior.

l4st1m0za (changes requested):
Great job! See the requested changes!

anastasiiavorobiova (approved):
Awesome work!

PR #1960, opened 2023-12-02
js_advanced_calculator

serjgrm:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+      return result + value;
+    },
+    subtract: function(result, value) {
+      return result - value;
+    },
+    multiply: function(result, value) {
+      return result * value;
+    },
+    divide: function(result, value) {
+      return result / value;
```
l4st1m0za, line 62: There should be a check for division by zero in the 'divide' function. In the current implementation, if 'value' is 0, it will return Infinity, which might break further calculations.

l4st1m0za (changes requested):
Great job! Consider the coment!

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #1964, opened 2023-12-06
Solution

RUBIC0N5545:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+      return this;
+    },
+    multiply(object, value) {
+      object.result *= value;
+
+      return this;
+    },
+    divide(object, value) {
+      object.result /= value;
```
anastasiiavorobiova, line 59: Consider checking division by zero, it's not allowed

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(operator, value) {
+      operator(this, value);
+
+      return this;
```
anastasiiavorobiova, line 71:
```suggestion

```

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(operator, value) {
+      operator(this, value);
```
anastasiiavorobiova, line 69:
```suggestion
      return  operator(this, value);
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #1966, opened 2023-12-07
calc soultion

steveforde:
updated solution 1

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -36,8 +36,62 @@
  *
  * @return {object}
  */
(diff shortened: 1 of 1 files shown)


anastasiia-tilikina (changes requested):
let's improve your code a bit.

you can create methods directly in your calculator:

```javascript
const calculator = {
  result: 0,
  
  add: (num) {
    // ...
    
    return this;
  }
}
```

lerastarynets (approved):
You're on the right track!

PR #1969, opened 2023-12-08
Solution by maks-n

ghost:


src/makeCalculator.js
```diff
@@ -37,7 +37,23 @@
  * @return {object}
  */
```
OleziO, line 38: Remove all comments from the code.

OleziO (approved):

PR #1970, opened 2023-12-08
js_advanse_calculator try#1

VolodymyrLiubyn:


src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
  */
 function makeCalculator() {
   // write code here
+  const calculatorObject = {
+    result: 0,
+
+    add(param) {
+      calculatorObject.result += param;
```
anastasiiavorobiova, line 45:
```suggestion
      this.result += param;
```

src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
...
+    },
+
+    subtract(param) {
+      calculatorObject.result -= param;
```
anastasiiavorobiova, line 49: Also, consider using this further in the code

src/makeCalculator.js
```diff
@@ -38,6 +38,40 @@
...
+    },
+
+    multiply(param) {
+      calculatorObject.result *= param;
+    },
+
+    divide(param) {
+      calculatorObject.result /= param;
```
anastasiiavorobiova, line 57: Consider implementing a check to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
VolodymyrLiubyn: Ok, if param = 0, divide(parap) would return infinity? or some error?

src/makeCalculator.js
VolodymyrLiubyn: Code doesn't work with "this", for methods "add", "subtract", "multiply", and "divide". When method "operate" calls some of them, it became "this", but it hasn't own property "result". I think so. For "operate" and "reset" "this" works, thank you.

PR #1971, opened 2023-12-09
Solution by GTXtab

GTXtab:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      obj.result -= value;
+    },
+
+    multiply(obj, value) {
+      obj.result *= value;
+    },
+
+    divide(obj, value) {
+      obj.result /= value;
+    },
```
etojeDenys, line 57: let's use `this` instead of the `obj`, it must improve the reusability of the function

etojeDenys (changes requested):

src/makeCalculator.js
GTXtab: image

etojeDenys: oh, sorry, you sent the obj as the argument

etojeDenys (approved):
lgtm

PR #1974, opened 2023-12-12
Solution

avakiel:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let sum = 0;
```
artem8746, line 40: create `result` property in object

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    },
+    subtract(num) {
+      sum -= num;
+
+      return this;
+    },
+    divide(num) {
+      sum /= num;
+
+      return this;
```
artem8746, line 61: why do you need this in methods `add`, `subtract`, `divide`, `multiply`, we use those methods with help of `operate` method, so we don't need to `return this` in this case

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+      sum *= num;
+
+      return this;
+    },
+    operate(callback, num) {
+      callback(num);
+
+      return this;
+    },
+    get result() {
```
artem8746, line 73: just create `result` property

src/makeCalculator.js
artem8746, line 40: you don't need this, just create `result` property

artem8746 (changes requested):
Good job!!
Just a few fixes needed

DanilWeda (approved):
Great! Approved! 
Thanks!

PR #1977, opened 2023-12-13
part 1

Manankin:
done !

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -38,29 +38,77 @@
  */
 function makeCalculator() {
   // write code here
-  let result = {
+  const calculator = {
     result: 0,
-    add: (value) => {
-      this.result += value;
+    add(value, result) {
+      return result + value;
     },
-    subtract: (value) => {
-      this.result -= value;
+
+    subtract(value, result) {
+      return result - value;
     },
-    multiply: (value) => {
-      this.result *= value;
+
+    multiply(value, result) {
+      return result * value;
     },
-    divide: (value) => {
-      this.result /= value;
+
+    divide(value, result) {
+      return result / value;
     },
-    reset: () => {
+
+    reset() {
       this.result = 0;
+
+      return this;
     },
-    operate: (fn, value) => {
-      this.fn(value);
-    }
+
+    operate(fn, value) {
+      this.result = fn(value, this.result);
+
+      return this;
+    },
   };
 
-  return result;
+  return calculator;
 }
+
+// function makeCalculator() {
+//   // write code here
+//   const res = {
+//     result: 0,
+//     add(value) {
+//       this.result += value;
+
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -38,6 +38,29 @@
  */
 function makeCalculator() {
   // write code here
+  let result = {
```
DanilWeda, line 41: Why do you use let in this case?

src/makeCalculator.js
```diff
@@ -38,6 +38,29 @@
...
+    },
+    subtract: (value) => {
+      this.result -= value;
+    },
+    multiply: (value) => {
+      this.result *= value;
+    },
+    divide: (value) => {
+      this.result /= value;
+    },
```
DanilWeda, line 54: You can add validation for zero-dividing

dothearts (changes requested):
Great! 
Check some comments!)

src/makeCalculator.js
Manankin: dividing by zero not describe i task conditions
(so I dont know what I must do in this case: return result without any action or something else)

src/makeCalculator.js
```diff
@@ -38,6 +38,29 @@
  */
 function makeCalculator() {
   // write code here
+  let result = {
+    result: 0,
+    add: (value) => {
+      this.result += value;
+    },
```
OleksandrRezanov, line 45: You dont need to use ':' and '=>'
add(value) { this.result += value; } - it works the same.

OleksandrRezanov (approved):

src/makeCalculator.js
```diff
@@ -38,6 +38,29 @@
...
+    subtract: (value) => {
+      this.result -= value;
+    },
+    multiply: (value) => {
+      this.result *= value;
+    },
```
olya-shyian, line 51: Use non-arrow functions for object methods (because  arrow function takes `this` from scope)

Please rewrite in this way:
image

This is a function that will  always receive meaningful `this` value from their caller

olya-shyian (changes requested):
Good gob! Try to fix comments =)

src/makeCalculator.js
```diff
@@ -38,6 +38,77 @@
...
+//     reset() {
+//       this.result = 0;
+
+//       return this;
+//     },
+//     operate(fn, value) {
+//       this.result = this.fn(value);
+
+//       return this;
+//     },
```
olya-shyian, line 108: don't forget to delete such commented code like this

olya-shyian (changes requested):
I wanted to approve but you have to remove commented code)
Let's do the last update

olya-shyian (approved):
Awesome!😉

PR #1979, opened 2023-12-15
add task solution

AgnieszkaWw:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return result - value;
+    },
+
+    multiply(result, value) {
+      return result * value;
+    },
+
+    divide(result, value) {
+      return result / value;
+    },
```
volodymyr-soltys97, line 69: It is necessary to provide for a check on 0, since it is not possible to divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Great work 👍 
Let's improve your code

PR #1980, opened 2023-12-15
add task solution

artem8746:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+  const calculator = {
+    result: 0,
+
+    operate(actionCallback, value) {
+      actionCallback(value, this);
+
+      return this;
+    },
+
+    add(value, obj = this) {
```
nabokok, line 51: Do we need a default value here?

nabokok (approved):
Good job! I left a small comment

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+      obj.result += value;
+    },
+
+    subtract(value, obj = this) {
+      obj.result -= value;
+    },
+
+    divide(value, obj = this) {
```
dinaminahi, line 59: Let's think of a case when value will be 0, we can't divide by 0

dinaminahi (changes requested):
Good job 👍  Let's only do one fix

src/makeCalculator.js
artem8746: if we're using operate method to call those functions - no, but if we want to use it without operate method it's needed

DanilWeda (approved):
Great! 
Approved!

PR #1983, opened 2023-12-15
solution

vSainiuk:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    multiply(value, result) {
+      return result * value;
+    },
+
+    divide(value, result) {
+      return result / value;
+    },
+
+    reset() {
+      this.result = 0;
```
artem8746, line 60: better create constant `RESET_DEFAULT_VALUE` or smth like this instead of hardcoded value(`0`)

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+      return this;
+    },
+
+    operate(operation, value) {
+      this.result = operation(value, this.result);
```
artem8746, line 66: think of a case when value type is not a number

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    subtract(value, result) {
+      return result - value;
+    },
+
+    multiply(value, result) {
+      return result * value;
+    },
+
+    divide(value, result) {
+      return result / value;
```
artem8746, line 56: think of a case when value is 0, we can't divide by 0

artem8746 (changes requested):
Good job!! 
Just a few fixes needed

src/makeCalculator.js
```diff
@@ -37,7 +37,79 @@
...
+//     reset: () => {
+//       this.result = RESET_DEFAULT_VALUE;
+
+//       return this;
+//     },
+//     operate: (operation, value) => {
+//       operation(value);
+
+//       return this;
+//     },
```
olya-shyian, line 113: please delete this commented code

src/makeCalculator.js
```diff
@@ -37,7 +37,79 @@
...
+    multiply(value, result) {
+      return result * value;
+    },
+
+    divide(value, result) {
+      if (value !== 0) {
+        return result / value;
+      }
+
+      throw Error;
```
olya-shyian, line 62: when you thow an error its good if you add some text about this error (in order to undersatand how to fix it)

in our case you can write: "You cannot divide by zero!"

olya-shyian (changes requested):
God job! I like it!😊
Just minor fixes needed

olya-shyian (changes requested):
Please do the last update (sorry I missed that)

Use a RESET_DEFAULT_VALUE as a value for result too:

image

PR #1986, opened 2023-12-18
add solution

ssaviak:


src/makeCalculator.js
```diff
@@ -36,8 +36,39 @@
...
+      return prevValue + number;
+    },
+
+    subtract(prevValue, number) {
+      return prevValue - number;
+    },
+
+    divide(prevValue, number) {
+      return prevValue / number;
+    },
```
vpashko, line 52: You can consider that division by zero is not allowed and return an error in such case.

src/makeCalculator.js
```diff
@@ -36,8 +36,39 @@
  *
  * @return {object}
  */
-function makeCalculator() {
-  // write code here
+function makeCalculator(callback, value) {
+  const resObj = {
```
vpashko, line 40: [NAMING] - Use proper names for variables. If function was called `makeFruit`, it should return `fruit`. (from checklist)
You can name your object, for example, 'calculator.'

vpashko (changes requested):
Great job! Let's improve your code.

vpashko (approved):
Good job! Approved

src/makeCalculator.js
```diff
@@ -36,8 +36,44 @@
...
+    subtract(prevValue, number) {
+      return prevValue - number;
+    },
+
+    divide(prevValue, number) {
+      if (number === 0) {
+        throw new Error('Dividing by zero! We keeping previous result');
+      } else {
+        return prevValue / number;
+      }
```
olya-shyian, line 56: you can remove this `else` and just return

src/makeCalculator.js
```diff
@@ -36,8 +36,44 @@
...
+    },
+
+    multiply(prevValue, number) {
+      return prevValue * number;
+    },
+
+    reset() {
+      this.result = 0;
```
olya-shyian, line 64: this `0`  you can keep in const as a start value

don't forget to use it in `result` as well

olya-shyian (changes requested):
Good job! 🤗 Just minor updates needed

DanilWeda (approved):
Great!
Approved!

PR #1987, opened 2023-12-18
add solution

OleksandrRezanov:


src/makeCalculator.js
```diff
@@ -37,7 +37,64 @@
...
+          this.divide(operand);
+
+          break;
+
+        default:
+          break;
+      }
+
+      return this;
+    },
```
ssaviak, line 70: there is unnecessary code. if you will have two parameters in each method and use callback in your operade method - the code becomes short.

src/makeCalculator.js
```diff
@@ -37,7 +37,64 @@
...
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
+      if (number === 0) {
+        return `Error! Deviding by zero!`;
```
ssaviak, line 86: spelling mistake [Deviding] = > [Dividing]

ssaviak (changes requested):
Good job! Also, remove unnecessary switch block, and use callback on whole power:)

olya-shyian (approved):
Marvelous work!😊

PR #1994, opened 2023-12-23
f

p1zzat1me:


src/makeCalculator.js
```diff
@@ -38,6 +38,36 @@
...
+    },
+    subtract: function(number) {
+      this.result -= number
+    },
+    multiply: function(number) {
+      this.result *= number
+    },
+    divide: function(number) {
+      this.result /= number
+    },
```
volodymyr-soltys97, line 60: It is necessary to provide for a check on 0, because according to the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Also add an actual title to the PR header instead of 'f'
image

Andriy-Shymkiv (approved):

PR #1999, opened 2023-12-24
Add advanced calculator solution

Voloshynd:


src/makeCalculator.js
```diff
@@ -1,43 +1,36 @@
...
+      this.result += number;
+    },
+
+    subtract(number) {
+      this.result -= number;
+    },
+
+    divide(number) {
+      this.result /= number;
+    },
```
volodymyr-soltys97, line 16: You need to add a check by 0, because according to the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
Voloshynd: Done

anastasiiavorobiova (approved):
Awesome!

PR #2007, opened 2024-01-02
Solution

OksanaZderko:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    subtract(value) {
+      this.result -= value;
+    },
+
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      this.result /= value;
```
l4st1m0za, line 56: There is no check for division by zero. If value is 0, it will cause an error because division by zero is undefined.

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      callback.call(this, value);
```
l4st1m0za, line 66: There is no validation of the callback function. There should be a check if the callback is a function before it's called.

l4st1m0za (changes requested):
Some changes needed

anastasiiavorobiova (approved):
Awesome work!

PR #2008, opened 2024-01-02
solution

Arkadiuszx:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      return this;
+    },
+
+    subtract(result, number) {
+      return result - number;
+    },
+
+    divide(result, number) {
+      return result / number;
+    },
```
volodymyr-soltys97, line 65: You need to add a check for 0, because by the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #2010, opened 2024-01-03
calculator

marichkamt:


src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
Viktor-Kost, line 40: this variable is redundant, you can return an object immediately 
```suggestion
  return {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+    operate(action, number) {
+      this.result = action(number, this.result);
+
+      return this;
+    },
+
+    add(a, b) {
+      return a + b;
+    },
```
Viktor-Kost, line 50: add/subtract/multiply/divide methods should have only one param and work with `this`
```suggestion
    add(value) {
      this.result += value;

      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    subtract(a, b) {
+      return b - a;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return b / a;
```
Viktor-Kost, line 61: add a condition to check if param isn't a zero

Viktor-Kost (changes requested):

PR #2012, opened 2024-01-04
Solution

YevhenProtasov:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    add(result, number) {
+      return result + number;
+    },
```
Viktor-Kost, line 43: 1. you should work with one function parameter
2. use this
```suggestion
    add(value) {
      this.result += value;

      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    subtract(result, number) {
+      return result - number;
+    },
+
+    multiply(result, number) {
+      return result * number;
+    },
+
+    divide(result, number) {
+      return result / number;
```
Viktor-Kost, line 54: add a condition to forbid dividing by zero

Viktor-Kost (changes requested):

Viktor-Kost (approved):

PR #2013, opened 2024-01-04
Solution

olduma:


src/makeCalculator.js
```diff
@@ -37,7 +37,31 @@
...
+    result: 0,
+    add(x) {
+      this.result += x;
+    },
+    subtract(x) {
+      this.result -= x;
+    },
+    multiply(x) {
+      this.result *= x;
+    },
```
polosanya, line 50:
```suggestion
    add(x) {
      this.result += x;
    },
    
    subtract(x) {
      this.result -= x;
    },
    
    multiply(x) {
      this.result *= x;
    },
```
Would be great to add newlines between methods for better readability)

polosanya (approved):
Nice!

PR #2015, opened 2024-01-06
calculator

marichkamt:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      return result - number;
+    },
+
+    multiply(result, number) {
+      return result * number;
+    },
+
+    divide(result, number) {
+      return result / number;
+    },
```
volodymyr-soltys97, line 69: You need to add a check for 0, because by the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

SanyaBratashchuk (approved):

PR #2020, opened 2024-01-08
add task solutions

MrWhiteres:


src/makeCalculator.js
```diff
@@ -33,11 +33,56 @@
...
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  const add = num => {
+    result += num;
+
+    return makeCalculator();
```
l4st1m0za, line 44: Return of 'makeCalculator' in the 'add' method is incorrect. It creates a new calculator each time. You should return 'this' to enable chaining.

src/makeCalculator.js
```diff
@@ -33,11 +33,56 @@
...
+    result *= num;
+
+    return makeCalculator();
+  };
+
+  const divide = num => {
+    result /= num;
+
+    return makeCalculator();
+  };
```
l4st1m0za, line 63: Return of 'makeCalculator' in the 'divide' method is incorrect. It creates a new calculator each time. You should return 'this' to enable chaining.

src/makeCalculator.js
```diff
@@ -33,11 +33,56 @@
...
+    result -= num;
+
+    return makeCalculator();
+  };
+
+  const multiply = num => {
+    result *= num;
+
+    return makeCalculator();
+  };
```
l4st1m0za, line 57: Return of 'makeCalculator' in the 'multiply' method is incorrect. It creates a new calculator each time. You should return 'this' to enable chaining.

src/makeCalculator.js
```diff
@@ -33,11 +33,56 @@
...
+
+  const divide = num => {
+    result /= num;
```
l4st1m0za, line 60: There is no check for division by zero. Add validation to prevent undefined behavior.

src/makeCalculator.js
```diff
@@ -33,11 +33,56 @@
...
+
+    return makeCalculator();
+  };
+
+  const reset = () => {
+    result = 0;
+
+    return makeCalculator();
+  };
```
l4st1m0za, line 69: Return of 'makeCalculator' in the 'reset' method is incorrect. It creates a new calculator each time. You should return 'this' to enable chaining.

src/makeCalculator.js
```diff
@@ -33,11 +33,56 @@
...
+    result += num;
+
+    return makeCalculator();
+  };
+
+  const subtract = num => {
+    result -= num;
+
+    return makeCalculator();
+  };
```
l4st1m0za, line 51: Return of 'makeCalculator' in the 'subtract' method is incorrect. It creates a new calculator each time. You should return 'this' to enable chaining.

l4st1m0za (changes requested):
Good job, but changes needed!

src/makeCalculator.js
```diff
@@ -33,11 +33,63 @@
...
+
+  const multiply = num => {
+    result *= num;
+
+    return this;
+  };
+
+  const divide = num => {
+    if (num === 0) {
```
l4st1m0za, line 60: Dividing by zero is not handled correctly. It's better to throw an error instead of just returning 'this'.

src/makeCalculator.js
```diff
@@ -33,11 +33,63 @@
  *
  * calculator.result === -28
  *
- *
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
```
l4st1m0za, line 39: Here you need to make not just usual **let** but you need **this.result = callback(this.result, number);**

You have some missanderstandings in this task. I would recommend you to ask in FE chat!

src/makeCalculator.js
```diff
@@ -33,11 +33,63 @@
...
+
+  const add = num => {
+    result += num;
+
+    return this;
+  };
```
l4st1m0za, line 45: This functions should work with this, so you need to add to **this**, same with other cases

l4st1m0za (changes requested):

src/makeCalculator.js
MrWhiteres: Lack of sleep affected my work, fixed the code to the proper style

l4st1m0za (approved):
Good job!

PR #2022, opened 2024-01-10
solution

fessenko-tr:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      this.result -= value;
+    },
+
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      this.result /= value;
+    },
```
volodymyr-soltys97, line 57: You need to add a check for 0, because according to the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

lerastarynets (approved):
lgtm 🚀

PR #2023, opened 2024-01-10
create solution

AnastasiiaHombalevska:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(accum, number) {
+      return accum - number;
+    },
+
+    multiply(accum, number) {
+      return accum * number;
+    },
+
+    divide(accum, number) {
+      return accum / number;
```
anastasiiavorobiova, line 62: Consider implementing a check if the number is not a 0 to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(accum, number) {
+      return accum - number;
+    },
+
+    multiply(accum, number) {
+      return accum * number;
+    },
+
+    divide(accum, number) {
+      return number > 0 ? accum / number : false;
```
Viktor-Kost, line 62: it's better to throw an error if value is `0` not just `false`

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    operate(toDo, number) {
+      this.result = toDo(this.result, number);
+
+      return this;
+    },
+
+    add(accum, number) {
+      return accum + number;
+    },
```
Viktor-Kost, line 51: 1. you should work with one function parameter
2. use `this`

```suggestion
    add(value) {
      this.result += value;
      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+  };
+
+  return calculator;
```
Viktor-Kost, line 72: redundant, you can immediately return an object 
```suggestion
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
```
Viktor-Kost, line 40:
```suggestion
  return {
```

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #2025, opened 2024-01-10
finished task

kapesha:


src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(result, a) {
+      return result + a;
+    },
```
Viktor-Kost, line 45: 1. you should work with one function parameter
2. use this

```suggestion
    add(value) {
      this.result += value;
      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    subtract(result, a) {
+      return result - a;
+    },
+    multiply(result, a) {
+      return result * a;
+    },
+    divide(result, a) {
```
Viktor-Kost, line 52: add a condition to forbid dividing by zero

Viktor-Kost (changes requested):

Andriy-Shymkiv (approved):

PR #2030, opened 2024-01-16
Calculator

AnnaPshenychnykh:


src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: (number) => calculator.result + number,
+    subtract: (number) => calculator.result - number,
+    multiply: (number) => calculator.result * number,
+    divide: (number) => calculator.result / number,
```
etojeDenys, line 45: let's use `this` instead of the `calculator` to make it much more reusable

src/makeCalculator.js
```diff
@@ -37,7 +37,26 @@
...
+    operate: (callback, number) => {
+      calculator.result = callback(number);
```
etojeDenys, line 47: (don't forget to bind the callback after that)

etojeDenys (changes requested):

maxim2310 (approved):
Good work!!!

PR #2033, opened 2024-01-16
Add solution

Vasya564:


ChrisHryts (approved):

src/makeCalculator.js
```diff
@@ -38,6 +38,39 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, value) {
+      operation.bind(this)(value);
```
polosanya, line 67:
```suggestion
      operation.call(this, value);
```
Call/apply methods are more relevant here as you immidiately call the function
https://www.w3schools.com/js/js_function_call.asp

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment 😉

src/makeCalculator.js
Vasya564: Thanks for the advice!

PR #2037, opened 2024-01-18
solution v_1

Beliyandr:


src/makeCalculator.js
```diff
@@ -37,7 +37,72 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const allOperates = {
+
+    // змінна в якій буде зберігатись результат
```
loralevitska, line 42: Remove comment. Try to write obvious code with meaningful variable and function naming
Fix all cases please

src/makeCalculator.js
```diff
@@ -37,7 +37,72 @@
...
+    set changeResult(number) {
+      this.resultOperation = number;
+    },
+
+    // дана функція приймає (callback) і число (number) - для маніпуляцій з ним
+    operate(callback, number) {
+      this.changeResult = callback(this.resultOperation, number);
+
+      return this.thisObject;
+    },
```
loralevitska, line 69: For better understanding `this` I recommend you to rewrite operate (other code you need will change too): 

```suggestion
    operate(operation, value) {
      operation.call(this, value);

      return this;
    },
```

loralevitska (changes requested):
Almost done) Let's improve

src/makeCalculator.js
volodymyr-soltys97: Still not fixed from the previous review, need to fix it everywhere
image

src/makeCalculator.js
```diff
@@ -1,43 +1,68 @@
...
+      const result = prevValue * value;
+
+      return result;
+    },
+
+    divide(prevValue, value) {
+      const result = prevValue / value;
+
+      return result;
+    },
```
volodymyr-soltys97, line 55: It is necessary to provide for a check by 0, because according to the rules of mathematics, it is not possible to divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -1,43 +1,60 @@
...
+    multiply(prevValue, value) {
+      const result = prevValue * value;
+
+      return result;
+    },
+
+    divide(prevValue, value) {
+      let result = prevValue / value;
+
+      if (value === 0) {
```
anastasiiavorobiova, line 43: It's better to make a check before division

anastasiiavorobiova (approved):
Great work!

PR #2040, opened 2024-01-21
solution

ant-tur:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(number) {
+      this.result += number;
+
+      return this.result;
```
Viktor-Kost, line 46: you can just return `this` (add for other operation as well)
```suggestion
      return this;
```

src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    },
+
+    multiply(number) {
+      this.result *= number;
+
+      return this.result;
+    },
+
+    divide(number) {
```
Viktor-Kost, line 61: add a condition to forbid dividing by zero

Viktor-Kost (changes requested):

Viktor-Kost (changes requested):
.

Viktor-Kost (approved):

PR #2041, opened 2024-01-22
Dima gres feature

Greshch:
Please check it. Thank's

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -38,6 +38,42 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    // operate : (calback, arg) => {
+    //   calback(arg);
+    // },
+  };
+
+  calculator['operate'] = (calback, arg) => {
+    calback(arg);
+
+    return calculator;
+  };
+
+  calculator['add'] = (arg) => {
+    calculator.result += arg;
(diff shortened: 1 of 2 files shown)


src/makeCalculator.js
```diff
@@ -38,6 +38,42 @@
  */
 function makeCalculator() {
   // write code here
+  const calculator = {
+    result: 0,
+    // operate : (calback, arg) => {
+    //   calback(arg);
+    // },
```
etojeDenys, line 45: looks better to implement like this, create the methods inside the calculator object directly

src/makeCalculator.js
```diff
@@ -38,6 +38,42 @@
...
+  };
+
+  calculator['operate'] = (calback, arg) => {
+    calback(arg);
+
+    return calculator;
+  };
+
+  calculator['add'] = (arg) => {
+    calculator.result += arg;
```
etojeDenys, line 55: better to use `this` instead of calculator, it will improve the reusability

etojeDenys (changes requested):

src/makeCalculator.js
Greshch: I've just push commit 'second implementation'. Please check it. But I can't use 'this` instead calculator[property](

PR #2045, opened 2024-01-24
add advanced calculator solution

lbihun:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+    };
+
+  calculator.multiply
+    = function(n) {
+      calculator.result *= n;
+    };
+
+  calculator.divide
+    = function(n) {
+      calculator.result /= n;
```
anastasiiavorobiova, line 68: Consider implementing a check to avoid division by `0`

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
```diff
@@ -37,7 +37,50 @@
...
+    };
+
+  calculator.add
+    = function(n) {
+      calculator.result += n;
+    };
+
+  calculator.subtract
+    = function(n) {
+      calculator.result -= n;
```
etojeDenys, line 59: better to use in `this` function, it will improve the reusability

etojeDenys (changes requested):

src/makeCalculator.js
maxim2310: fix it in all places

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

sTorba24 (approved):
GJ 🚀

PR #2046, opened 2024-01-25
add task solution

AnastasiaChaikovskaya:
add task solution for js_advanced_calculator

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    operate: function(operation, number) {
+      operation.bind(this)(number);
```
TarasHoliuk, line 43: :+1: but you may use `.call` instead of `bind` if you want to call the function immediately
just FYI

TarasHoliuk (approved):
:heavy_check_mark:

PR #2047, opened 2024-01-26
Solution

ayasinska:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    multiply: function(num) {
+      this.result *= num;
+
+      return this;
+    },
+
+    divide: function(num) {
+      if (num !== 0) {
```
vlasiuksasha, line 62: In the divide function, there's no handling for the case when the number is 0. This will cause a division by zero error. Please handle this case properly, perhaps by throwing an error.

src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    reset: function() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate: function(callback, num) {
+      callback.call(this, num);
```
vlasiuksasha, line 76: In the operate function, there's no checking if the callback is a function. This could lead to a runtime error if a non-function is passed. Please add an appropriate check.

vlasiuksasha (changes requested):
check the comments

4ksandrey (approved):
Accepted, well done

PR #2049, opened 2024-01-26
task solution

Poviakalo:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      return value - n;
+    },
+
+    multiply(value = 0, n = 1) {
+      return value * n;
+    },
+
+    divide(value = 0, n = 1) {
+      return value / n;
+    },
```
volodymyr-soltys97, line 57: You need to add a check by 0, because according to the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

sTorba24 (approved):
GJ 🚀

PR #2051, opened 2024-01-28
Solution

Yehorf21:
Completed the task

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+
+      return this;
+    },
+    subtract(number) {
+      obj.result -= number;
+
+      return this;
+    },
+    divide(number) {
+      obj.result /= number;
```
anastasiiavorobiova, line 58: Consider checking if the num is not 0 to avoid division by 0

anastasiiavorobiova (approved):
Great work!

PR #2053, opened 2024-01-29
solution

VasulenkoIllia:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
```
Viktor-Kost, line 40: you can return immediately 
```suggestion
  return {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+    operate(operation, number) {
+      operation(number);
+
+      return this;
+    },
+  };
 
+  calculator.result = 0;
+
+  return calculator;
```
Viktor-Kost, line 67:
```suggestion
```

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      return this;
+    },
+    operate(operation, number) {
+      operation(number);
+
+      return this;
+    },
+  };
 
+  calculator.result = 0;
```
Viktor-Kost, line 65: move inside an object like `result: 0,`

src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  const calculator = {
+    add(number) {
+      calculator.result += number;
```
Viktor-Kost, line 42: use `this`. apply for other operations as well
```suggestion
      this.result += number;

      return this.result;
```

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #2061, opened 2024-01-31
add task solution

Roman-Bilichenko:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      return num1 - num2;
+    },
+
+    divide(num1, num2) {
+      return num1 / num2;
+    },
+
+    multiply(num1, num2) {
+      return num1 * num2;
+    },
```
Esceype, line 63:
```suggestion
    add(currentValue, number) {
      return currentValue+ number;
    },

    subtract(num1, num2) {
      return num1 - num2;
    },

    divide(num1, num2) {
      return num1 / num2;
    },

    multiply(num1, num2) {
      return num1 * num2;
    },
```

Naming of arguments and variables is important, num1 and num2 are not informative enough and not quite correct, it should be currentValue, and number

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      return num1 + num2;
+    },
+
+    subtract(num1, num2) {
+      return num1 - num2;
+    },
+
+    divide(num1, num2) {
+      return num1 / num2;
+    },
```
Esceype, line 59: You need to process the variant if the number to be divided by is zero since you cannot divide by zero.

Esceype (changes requested):
GJ!
Read the following comments

lerastarynets (approved):
lgtm 🚀

PR #2063, opened 2024-02-01
add a solution

ShvetsOlha:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    subtract(number) {
+      this.result -= number;
+
+      return this.result;
+    },
+
+    divide(number) {
+      this.result /= number;
```
maxim2310, line 54: need to add validation if number will be 0

maxim2310 (changes requested):

src/makeCalculator.js
lerastarynets: not fixed

lerastarynets (changes requested):
Pls fix prev comments

OleziO (approved):

PR #2068, opened 2024-02-03
Solution

Galers:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+      this.result -= a;
+    },
+
+    multiply(a) {
+      this.result *= a;
+    },
+
+    divide(a) {
+      this.result /= a;
+    },
```
volodymyr-soltys97, line 69: You need to add a check on 0, because you cannot divide by 0, according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Great work 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #2069, opened 2024-02-04
add task solution

DenLysiak:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      sum.result -= number;
+    },
+
+    multiply(number) {
+      sum.result *= number;
+    },
+
+    divide(number) {
+      sum.result /= number;
+    },
```
volodymyr-soltys97, line 57: You need to add a check by 0, because according to the rules of mathematics, you cannot divide by 0

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+
+    operate(callBack, argument) {
+      callBack(argument);
```
volodymyr-soltys97, line 60:
```suggestion
      callBack.call(this, argument);
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

etojeDenys (approved):
lgtm

PR #2070, opened 2024-02-05
finish

Zubkovskiy:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    subtract(num) {
+      this.result -= num;
+    },
+
+    multiply(num) {
+      this.result *= num;
+    },
+
+    divide(num) {
+      this.result /= num;
```
l4st1m0za, line 56: Division by zero is not handled. It would lead to Infinity or -Infinity, which would be a bug in the application.

l4st1m0za (changes requested):
Good job, but consider the comment!

anastasiiavorobiova (approved):
Great work!

PR #2072, opened 2024-02-06
create calculator

denisbielykh:


src/makeCalculator.js
```diff
@@ -37,7 +37,38 @@
...
+      return result - value;
+    },
+
+    multiply(result, value) {
+      return result * value;
+    },
+
+    divide(result, value) {
+      return result / value;
+    },
```
lerastarynets, line 68: pls add case with dividing to 0, and return an error because it's impossible

lerastarynets (changes requested):

src/makeCalculator.js
denisbielykh: console.log('Error'); линтер не пропускает. Добавил просто проверку на 0

etojeDenys (approved):
good job

PR #2074, opened 2024-02-06
Solution

Alexraiv:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(a, b) {
+      return a + b;
+    },
```
Viktor-Kost, line 57: you can update your operations to work with current value and `this`. Apply for operations below as well
```suggestion
    add(a) {
      this.result += a;
      return this;
    },
```

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
```
Viktor-Kost, line 67: add a condition to forbid dividing by zero

Viktor-Kost (approved):
Nice!

PR #2078, opened 2024-02-07
Solution

wasik07:


src/makeCalculator.js
```diff
@@ -36,8 +36,61 @@
...
+      return this;
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(value, number) {
```
etojeDenys, line 73: image
you need to take the callback as the parameter
image

etojeDenys (changes requested):
feel free to tag me in the chat and ask for some help there

src/makeCalculator.js
```diff
@@ -36,8 +36,41 @@
...
+      calculator.result -= figure;
+    },
+
+    multiply(figure) {
+      calculator.result *= figure;
+    },
+
+    divide(figure) {
+      calculator.result /= figure;
+    },
```
lerastarynets, line 64: consider case when figure is 0 and return an error, because we can't divide to 0

lerastarynets (changes requested):
Almost there, just one problem left)

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2084, opened 2024-02-09
Solution

in7264:


src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let output = 0;
+
+  return {
+    add(x) {
```
volodymyr-soltys97, line 43:
```suggestion
  let output = 0;

  return {
    result: 0,
    
    add(x) {
```

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+      output *= x;
+
+      return this;
+    },
+
+    divide(x) {
+      output /= x;
+
+      return this;
+    },
```
volodymyr-soltys97, line 65: You need to add a check for 0, because by the rules of mathematics, you cannot divide by 0

src/makeCalculator.js
```diff
@@ -37,7 +37,49 @@
...
+
+    subtract(x) {
+      output -= x;
+
+      return this;
+    },
+
+    multiply(x) {
+      output *= x;
```
volodymyr-soltys97, line 56: Fix it everywhere
```suggestion
      this.result *= x;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
in7264: `function makeCalculator() {
  let output = 0;

  return {
    result: 0,

    add(x) {
      this.result += x;

      return this;
    },

    subtract(x) {
      this.result -= x;

      return this;
    },

    multiply(x) {
      this.result *= x;

      return this;
    },

    divide(x) {
      if (x !== 0) {
        this.result /= x;
      } else {
        return 'Division by zero!';
      }

      return this;
    },

    reset() {
      this.result = 0;

      return this;
    },

    operate(operation, operand) {
      operation(operand);

      return this;
    },

    get result() {
      return this.result;
    },
  };
}`
I did everything as you said, but jest gives me an error
```
sum > 'operate' should correctly add a single value
-----
TypeError: Cannot read properties of undefined (reading 'result')
```
I tried giving different names to the variables but it didn't help

etojeDenys (approved):

PR #2085, opened 2024-02-10
calculator

OMedvid:


Andriy-Shymkiv (changes requested):
please fix your code according to [checklist](https://github.com/mate-academy/js_advanced_calculator/blob/master/checklist.md)

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+
+      return this;
+    },
+    multiply(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide(num) {
+      this.result /= num;
```
volodymyr-soltys97, line 59: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/makeCalculator.js
```diff
@@ -37,7 +37,44 @@
...
+      return this;
+    },
+    multiply(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide(num) {
+      if (num === 0) {
+        return this;
```
volodymyr-soltys97, line 60:
```suggestion
        return throw new Error('You cannot divide by 0');
```

volodymyr-soltys97 (approved):
Good job 👍

PR #2089, opened 2024-02-12
implement solution

Aleksandra-hub:


src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    multiply(num) {
+      calculator1.result *= num; ;
+    },
+    divide(num) {
+      calculator1.result /= num;
+    },
+    reset() {
+      calculator1.result = 0;
+
+      return calculator1;
```
VitaliyBondarenko1982, line 57: use this in all methods instead calculator1

src/makeCalculator.js
```diff
@@ -37,7 +37,34 @@
...
+    },
+    operate(operand, number) {
+      operand(number);
```
VitaliyBondarenko1982, line 60:
```suggestion
      operand.call(this, number);
```
here can be problem with this in operand method, when you will use this in methods. so you can set this directly with call method.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments.

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #2092, opened 2024-02-13
Solution

mfcastilho:


jv-aquino (changes requested):
Great code overall with good readability and following the best practices (like using function methods directly)! I would just ask you to shorten the code a little to make it more concise -but don't sacrifice readability for that. Anyways, great job!

PR #2094, opened 2024-02-14
Solution

KachVl:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(number) {
+      calculator.result += number;
```
etojeDenys, line 44: better to use `this` instead of the `calculator` here, it will improve the reusability

etojeDenys (approved):
good job

PR #2096, opened 2024-02-14
Solution

kovaltar:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
```
maxim2310, line 68: need to add validation if b === 0

maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed

src/makeCalculator.js
kovaltar: done

Andriy-Shymkiv (approved):

PR #2097, opened 2024-02-14
created advanced calculator

Zubyk-Yaroslav:
Hello my prev version looked like

`function makeCalculator() {
  return {
    result: 0,

    operate(callbackOperation, number) {
      this.result = callbackOperation(number);

      return this;
    },

    add(a) {
      return this.result + a;
    },

   ...

    reset() {
      this.result = 0;

      return this;
    },
  };
}`

but it didn't work. Can you explain please why the current version of the function is working, but version above doesn't work?
Thanks!

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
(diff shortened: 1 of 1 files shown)


etojeDenys (approved):
lgtm

PR #2101, opened 2024-02-18
Add solution

ViktoriiaRepo:


src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
-}
+  return {
+    result: 0,
+    add: function(number) {
```
polosanya, line 42: image

src/makeCalculator.js
```diff
@@ -37,7 +37,43 @@
...
+      this.result += number;
+
+      return this;
```
polosanya, line 45: There is no need to return `this` for each method. Only reset and operate methods should be called in chain

polosanya (changes requested):
Almost done! Don't forget to check yourself with the checklist

Moroz-Dmytro (approved):
Looks good 👍

PR #2103, opened 2024-02-19
calculator done

anastasia1383:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      this.result += num;
+
+      return this;
```
ViktoriiaRepo, line 46: There is no need to return this for each method. Only reset and operate methods should be called in chain

ViktoriiaRepo (approved):
Good job

Moroz-Dmytro (approved):

PR #2104, opened 2024-02-19
complete

TideTwo:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    operate(callback, value = null) {
```
Davwidos, line 42: Not sure why null as deafult value. Null still changes value of result in multiply and divide method. But becuause in this task we did not consider cases where number is not provided i think it is ok.

Davwidos (approved):
Nice job

choeqq (approved):

PR #2106, opened 2024-02-20
 created function makeCalculator()

YarynaPuhach:


src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+
+      return calculator.result;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return calculator;
+    },
+    operate(callback, num) {
+      calculator.result = callback(calculator.result, num);
```
TideTwo, line 67: You could use `callback.call(this, num);` when invoking methods like `add` inside `operation`
This gives context to the methods called by `operation` and lets them use `this` in reference to the full `calculator` object inside each of its methods, for example: `this.result = 0;` for the reset method. We are not taught `.call` or other object methods before this task, so I'm marking this solution as approved.

TideTwo (approved):

src/makeCalculator.js
choeqq: yup, I agree 😄

choeqq (changes requested):
as comment above ⬆️

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    add(a, b) {
```
asakevych, line 41: no need to accept 2 arguments, you can only accept 1 and use this.result

src/makeCalculator.js
```diff
@@ -37,7 +37,41 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    operate(callback, num) {
+      this.result = callback.call(this, this.result, num);
+
+      return this;
+    },
```
asakevych, line 70: no need to path this.result to function as argument, you have the access to the context inside the function when you call it with `.call` and your context

asakevych (changes requested):
Good job :rocket: small change is needed

asakevych (approved):
Nice! :fire:

PR #2107, opened 2024-02-20
add task solution

ShaoLei2:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+  function subtract(value) {
+    result -= value;
+  }
+
+  function multiply(value) {
+    result *= value;
+  }
+
+  function divide(value) {
+    result /= value;
```
l4st1m0za, line 55: There is no check for divide by zero. Please handle this case because division by zero is undefined.

l4st1m0za (changes requested):
Good job, some changes needed!

l4st1m0za (approved):
Good job!

PR #2116, opened 2024-02-25
Solved task --- have a clarification

natalia-matash:
I passed all tests, but i think my solution is a bit weak. How i can improve my code? If it is possible, please give me a hint and i will change it


etojeDenys (approved):
lgtm

PR #2120, opened 2024-02-26
Develop

getchard:


src/makeCalculator.js
```diff
@@ -1,43 +1,17 @@
...
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(a) { calculator.result += a; return calculator; },
+    subtract(a) { calculator.result -= a; return calculator; },
+    multiply(a) { calculator.result *= a; return calculator; },
+    divide(a) { calculator.result /= a; return calculator; },
+    reset() { calculator.result = 0; return calculator; },
+    operate(callback, number) { return callback.call(calculator, number); }
```
VitaliyBondarenko1982, line 11:
```suggestion
    add(a) { 
      this.result += a; 
      
      return this;
    },
    subtract(a) { 
      this.result -= a;
      
      return this;
     },
    multiply(a) { 
      this.result *= a; 
      
      return this; 
    },
    divide(a) { 
      this.result /= a;
      
       return this; 
     },
    reset() {
      this.result = 0;
      return this;
     },
    operate(callback, number) {
       return callback.call(this, number); 
    }
```

VitaliyBondarenko1982 (approved):
Approve, as was problem with node versions locally.
But please check comment: use this instead calculator

PR #2126, opened 2024-02-27
Develop

Maksym-Atanasov:


src/makeCalculator.js
```diff
@@ -37,7 +37,35 @@
...
+    subtract (value, currentResult) {
+      return currentResult - value;
+    },
+
+    multiply (value, currentResult) {
+      return currentResult * value;
+    },
+
+    divide (value, currentResult) {
+      return currentResult / value;
```
maxim2310, line 61: need to add validation when value === 0

maxim2310 (changes requested):

src/makeCalculator.js
```diff
@@ -37,7 +37,40 @@
...
+    multiply (value, currentResult) {
+      return currentResult * value;
+    },
+
+    divide (value, currentResult) {
+      if (value === 0) {
+        return currentResult / 1;
+      } else {
+        return currentResult / value;
+      }
```
etojeDenys, line 65:
```suggestion
      }
      
      return currentResult / value;
      
```

etojeDenys (approved):
lgtm

PR #2129, opened 2024-02-29
Implement calculator

deezhizyu:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    // Operations
+    add(obj, value) {
+      obj.result += value;
+    },
```
lerastarynets, line 59:
```suggestion
    add(value) {
      this.result += value;
    },
```
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+      obj.result -= value;
+    },
+
+    multiply(obj, value) {
+      obj.result *= value;
+    },
+
+    divide(obj, value) {
+      obj.result /= value;
+    },
```
lerastarynets, line 71: pls consider case of division to 0 and throw an error in this case

lerastarynets (changes requested):

src/makeCalculator.js
deezhizyu: doesn't work, i tried
image

lerastarynets: this is probably because the implementation of operate function is also wrong

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
-  // write code here
+  return {
+    result: 0,
+
+    // Main methods
+    operate(callback, value) {
+      callback(this, value);
+
+      return this;
+    },
```
lerastarynets, line 48:
```suggestion
     operate: function(callback, num) {
      callback.call(this, num);

      return this;
    },
```

lerastarynets (changes requested):
Pls take my comments into account

src/makeCalculator.js
deezhizyu: [CODE KNOWLEDGE] - if you creating a method in the object, you don't need to use function keyword, use shortcut instead.
BAD EXAMPLE:

 methodName: function() {
 // your code
},
GOOD EXAMPLE:

 methodName() {
 // your code
},

deezhizyu: Thanks, with callback it works! Strange that it isn't in course

anastasiiavorobiova (approved):
Awesome work!

PR #2138, opened 2024-03-01
Solution

lLiashko:


src/makeCalculator.js
```diff
@@ -38,6 +38,37 @@
...
+
+      return this;
+    },
+    multiply(num) {
+      this.result *= num;
+
+      return this;
+    },
+    divide(num) {
+      this.result /= num;
```
anastasiiavorobiova, line 59: Consider implementing a check to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
anastasiiavorobiova: Not fixed

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #2141, opened 2024-03-03
add task solution

silchenko-arsen:


src/makeCalculator.js
```diff
@@ -37,7 +37,47 @@
...
+    },
+
+    multiply: (num) => {
+      calculator.result *= num;
+
+      return calculator;
+    },
+
+    divide: (num) => {
+      calculator.result /= num;
```
l4st1m0za, line 74: There is no check for division by zero. This will lead to Infinity result or even crash.

l4st1m0za (changes requested):
Good job, but one meaningful change is needed

l4st1m0za (approved):
Great job!

PR #2145, opened 2024-03-05
Make Calculator

genakorytov:


src/makeCalculator.js
```diff
@@ -37,7 +37,42 @@
...
+
+      return this;
+    },
+    divide(number) {
+      this.result = this.result / number;
+
+      return this;
+    },
+    reset() {
+      this.result = this.result * 0;
```
etojeDenys, line 69:
```suggestion
      this.result *= 0;
```
we can simplify all functions like that

etojeDenys (approved):
good job

PR #2149, opened 2024-03-08
Solution

pKaterynaa:


src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
...
+    multiply(a) {
+      this.result *= a;
+
+      return this;
+    },
+    divide(a) {
+      this.result /= a;
+
+      return this;
+    },
```
volodymyr-soltys97, line 61: You need to add a check for 0, because by the rules of mathematics, you cannot divide by 0

src/makeCalculator.js
```diff
@@ -37,7 +37,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(a) {
```
volodymyr-soltys97, line 42: Use a more descriptive name for value instead of `a`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

lerastarynets (approved):
Keep up the great work!

PR #2151, opened 2024-03-09
Advanced_Calculator_Solution

FEARParanoia:


src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    },
+
+    reset() {
+      resultingObject.result = 0;
+
+      return resultingObject;
+    },
+
+    operate(callback, number) {
+      resultingObject.result = callback(resultingObject.result, number);
```
volodymyr-soltys97, line 66:
```suggestion
      callback.call(this, number);
```

src/makeCalculator.js
```diff
@@ -37,7 +37,39 @@
...
+    subtract(a, b = 0) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
```
volodymyr-soltys97, line 56: You need to add a check by 0, because according to the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

etojeDenys (approved):
lgtm

PR #2152, opened 2024-03-09
game over

true-binar-White-Rabbit:


src/makeCalculator.js
```diff
@@ -37,7 +37,45 @@
...
+  const calc = {
+    result: 0,
+    operate(callAction, num) {
+      callAction(num);
+
+      return calc;
+    },
+
+    add(prop) {
+      calc.result += prop;
```
etojeDenys, line 49: better to use `this` here instead of the calc, it will improve the reusability

etojeDenys (approved):
good job

src/makeCalculator.js
true-binar-White-Rabbit: try to use 'this' here instead calc and after we will discuss why is no working properly

PR #2153, opened 2024-03-11
solution

ilyaivasyk:


src/makeCalculator.js
```diff
@@ -36,8 +36,37 @@
...
  * @return {object}
  */
-function makeCalculator() {
-  // write code here
+function makeCalculator(calculator) {
+  const calc = {
+    result: 0,
+
+    add(num) {
+      calc.result += num;
```
etojeDenys, line 44: better to use `this` instead of calc, it will help to improve the reusability

etojeDenys (approved):
good job

PR #2162, opened 2024-03-14
solution

taras-nakonechnyi:


src/makeCalculator.js
```diff
@@ -37,7 +37,33 @@
...
+      return calculator;
+    },
+    add(value) {
+      calculator.result += value;
+    },
+    subtract(value) {
+      calculator.result -= value;
+    },
+    divide(value) {
+      calculator.result /= value;
```
lerastarynets, line 54: pls consider case, wiith division to 0, in this case an error should be shown

lerastarynets (changes requested):

anastasiiavorobiova (approved):
GJ!

PR #2166, opened 2024-03-15
Solution

aMtiIV:


src/makeCalculator.js
```diff
@@ -4,7 +4,26 @@
...
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
+    },
+
+    add: (a, b) => a + b,
```
BogdanMaliuta, line 22: why a+b. not result+value?

src/makeCalculator.js
```diff
@@ -4,7 +4,26 @@
...
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => a / b,
```
BogdanMaliuta, line 25: it means that I can divide in your calc by 0?

BogdanMaliuta (changes requested):
GJ, but try to improve your code

src/makeCalculator.js
```diff
@@ -4,7 +4,32 @@
...
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
+
+      return this;
+    },
+
+    add: (result, value) => result + value,
```
BogdanMaliuta, line 22: i think better to take in your add method only value, ang get result from this

src/makeCalculator.js
```diff
@@ -4,7 +4,32 @@
...
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(operation, number) {
+      this.result = operation(this.result, number);
```
BogdanMaliuta, line 17: here you need to make .call for operation, and pass this, not this,result

BogdanMaliuta (changes requested):
Some additional yhigs to change

BogdanMaliuta (approved):
Nice!

PR #2174, opened 2024-03-19
created

DmitriyChernega:


src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+      return number1 - number2;
+    },
+
+    multiply(number1, number2) {
+      return number1 * number2;
+    },
+
+    divide(result, number2) {
+      return result / number2;
+    },
```
lerastarynets, line 24: consider case with devision to 0, and throw errors in this case

lerastarynets (changes requested):

src/makeCalculator.js
```diff
@@ -4,7 +4,42 @@
...
+      return number1 - number2;
+    },
+
+    multiply(number1, number2) {
+      return number1 * number2;
+    },
+
+    divide(result, number2) {
+      if (number2 !== 0) {
+      }
```
etojeDenys, line 24: why do you need it, looks redundant

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2178, opened 2024-03-21
add task solution

AlexandeeerG:
Make Calculator
Another calculator. Now the task is more difficult.

Create a makeCalculator function that returns an object that has the following fields:

Methods: add, subtract, multiply, divide, reset, operate.
The result property is initially 0.
How the calculator will work:

Each operate call takes a callback and a number and sets the appropriate value to the result property.
The reset method resets result value to 0.
add, subtract, multiply, divide are passed as callbacks to operate method.
The operate and reset methods can be called in a chain.
Example:

const calculator = makeCalculator();

calculator.operate(calculator.add, 21)

console.log(calculator.result); // 21

calculator.reset()
console.log(calculator.result); // 0

calculator
  .operate(calculator.add, 10)
  .reset()
  .operate(calculator.subtract, 20)
  .operate(calculator.divide, 5)
  .operate(calculator.multiply, 7)

console.log(calculator.result); // -28

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -4,7 +4,57 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      const number = this;
+
+      number.result += num;
+
+      return number;
+    },
+
+    subtract(num) {
+      const number = this;
+
+      number.result -= num;
+
+      return number;
+    },
+
+    multiply(num) {
+      const number = this;
+
+      number.result *= num;
+
+      return number;
+    },
+
+    divide(num) {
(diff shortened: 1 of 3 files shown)


src/makeCalculator.js
```diff
@@ -4,7 +4,57 @@
...
-  // write code here
+  return {
+    result: 0,
+
+    add(num) {
+      const number = this;
+
+      number.result += num;
+
+      return number;
```
etojeDenys, line 15:
```suggestion
      this.result += num;

      return this;
```
just use `this` directly

etojeDenys (changes requested):
fix it everywhere

PR #2182, opened 2024-03-26
task solution

Sywuxy:


src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+
+    add: (value) => {
```
volodymyr-soltys97, line 10: Will be better if you fix it everywhere
image
```suggestion
    add(value) {
```

src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+      return (obj.result += value);
+    },
+
+    subtract: (value) => {
+      return (obj.result -= value);
```
volodymyr-soltys97, line 15: Fix it everywhere
```suggestion
      this.result -= value;
      
      return this;
```

src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    },
+
+    multiply: (value) => {
+      return (obj.result *= value);
+    },
+
+    divide: (value) => {
+      return (obj.result /= value);
```
volodymyr-soltys97, line 23: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/makeCalculator.js
```diff
@@ -4,7 +4,49 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const obj = {
+    result: 0,
+
+    add(value) {
+      obj.result += value;
```
DarkMistyRoom, line 11:
```suggestion
      this.result += value;
```

DarkMistyRoom: the same for the other methods

DarkMistyRoom (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2185, opened 2024-03-27
Solution

eduvinagre:


src/makeCalculator.js
```diff
@@ -4,7 +4,46 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calc = {
+    result: 0,
+
+    add(input) {
+      this.result += input;
+
+      return this.result;
```
jv-aquino, line 13: you could simplify this return here using a single line... can you find the solution? (remember about operator precedence and that return always computes the operators before returning anything)

jv-aquino (changes requested):
Nice job Edu! Your code is perfect in terms of logic and syntax, this is awesome for real! There's a small challenge that I left for you so you can better understand js operator precedence and return statements, if you don't want to do it that's ok, just request review again that I'll approve it. Anyways, hope you like it!

Some reference of op precedence:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence

jv-aquino (approved):
Great job Edu! Your solution was already awesome, but now it's top tier!! I'll tell you more  about it on the chat

PR #2186, opened 2024-03-29
VB solution

vbuldenko:


src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    multiply: function (number) {
+      this.result *= number;
+
+      return this;
+    },
+    divide: function (number) {
+      this.result /= number;
+
+      return this;
+    },
```
volodymyr-soltys97, line 28: You need to add a check for 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

PR #2189, opened 2024-04-01
Develop

ArtemZakhar:
Would appreciate it if you show an alternative variant of the "operate" method implementation.
Have a nice day!)

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -25,9 +25,11 @@ function makeCalculator() {
 
     reset() {
       this.num = 0;
+
+      return this;
     },
 
-    result() {
+    get result() {
       return this.num;
     },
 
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -4,7 +4,43 @@
...
+
+      return this;
+    },
+
+    get result() {
+      return this.num;
+    },
+
+    operate(callback, value) {
+      const func = callback.bind(this, value);
```
mrkovaliv, line 37: just 

```
            callback.call(this, value);
            return this;
```

mrkovaliv (approved):
Nice work 👍 👍 👍

src/makeCalculator.js
ArtemZakhar: the better approach I saw in Artem decision latter. 

callback(this.num, value);

ArtemZakhar: Thank you!)

PR #2192, opened 2024-04-02
add task solution

PrytulaAlexandr:


src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
...
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      this.result += value;
+
+      return this;
```
alexander-ignatow, line 13: add / return / multiply / subtract does not have to return `this`

alexander-ignatow (approved):
Good 👍 , but see my comment

PR #2200, opened 2024-04-06
make calculator

VvynnykV:
I dont understand why we should use " method call()" in this, as we dont change 'THIS' in the call(), but just put the same 'THIS'. Why we cant just do like this: 
'operate(callback, num) {
      callback(num);

      return this;
    },'


VitaliyBondarenko1982 (approved):
Good job.
About your question.
callback as argument of operate method will lose this, so we need to set it directly.

PR #2201, opened 2024-04-07
Solution

alshust:


src/makeCalculator.js
```diff
@@ -37,7 +37,28 @@
...
+    operate(action, number) {
+      this.result = action(this.result, number);
+
+      return this;
+    },
+
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => a / b,
```
anastasiiavorobiova, line 52: Consider implementing a check to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
```diff
@@ -37,7 +37,28 @@
...
+    operate(action, number) {
+      this.result = action(this.result, number);
+
+      return this;
+    },
+
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => (b > 0 ? a / b : 'Cann`t be devided by 0'),
```
anastasiiavorobiova, line 52: The division by subzero numbers should be possible. Also, consider returning 0 instead of the string or throwing an error

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Great work!

PR #2202, opened 2024-04-08
Solution

yuliaPel:


src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, a) {
+      callback.call(this, a);
```
FaiHamid, line 33: You also can write "callback(a);", it will work the same way )

FaiHamid (approved):
Well done)))

PR #2212, opened 2024-04-12
add task solution

danzaris:


src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
...
-  // write code here
+  const calculator = {
+    result: 0,
+    operate(callback, operand) {
+      callback(operand);
+
+      return calculator;
+    },
+    add(num) {
+      calculator.result += num;
```
etojeDenys, line 15: better to use `this` instead of the calculator, it will improve the reusability
```suggestion
    add(num) {
      this.result += num;
```

etojeDenys (approved):
lgtm

PR #2214, opened 2024-04-13
Added solution

vadym-valchak:


src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  // let result = 0;
```
VitaliyBondarenko1982, line 7:
```suggestion
```
do not keep redundant commented code

src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      this.result = callback.call(this, value);
```
VitaliyBondarenko1982, line 43:
```suggestion
      callback.call(this, value);
```
looks like redundant to reassign this.result.
And in general no need to return this.result from each operate method. enough just calculate this.result.

VitaliyBondarenko1982 (approved):
Good job.

PR #2216, opened 2024-04-16
Solution

hardixxxxx:


anastasiiavorobiova (changes requested):
Great work! Consider implementing a check to avoid division by 0

lerastarynets (approved):
You're on the right track!

PR #2218, opened 2024-04-16
Develop

MarkGodnyi:


src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    subtract(num) {
+      calculator.result -= num;
+    },
+
+    multiply(num) {
+      calculator.result *= num;
+    },
+
+    divide(num) {
+      calculator.result /= num;
```
maxim2310, line 23: need add validation for num === 0

maxim2310 (changes requested):

src/makeCalculator.js
MarkGodnyi: What should I assign to the result in that case?

anastasiiavorobiova (approved):
Great work!

PR #2219, opened 2024-04-16
Advanced calculator

AvramenkoMarina:
Create a `makeCalculator` function that returns an object that
has the following fields:
- Methods: `add`, `subtract`, `multiply`, `divide`, `reset`, `operate`.
- The `result` property is initially 0.


etojeDenys (approved):
lgtm

PR #2221, opened 2024-04-17
add solution

Ellworts:
<3

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -4,7 +4,38 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculatorObject = {
+    result: 0,
+    operate(operator, value) {
+      this.result = operator(this.result, value);
+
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -4,7 +4,38 @@
...
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
+      return a / b;
```
anastasiiavorobiova, line 28: Consider implementing a check to avoid division by 0

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/makeCalculator.js
Ellworts: @anastasiiavorobiova done ✅

etojeDenys (approved):
good job

PR #2233, opened 2024-04-27
Solution

danil-maksimenko:


src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+      return this.result - number;
+    },
+
+    multiply(number) {
+      return this.result * number;
+    },
+
+    divide(number) {
+      return this.result / number;
+    },
```
volodymyr-soltys97, line 22: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great work 🔥

PR #2234, opened 2024-04-27
js_advanced_calculator

DmitryVolchokFE:


src/makeCalculator.js
```diff
@@ -4,7 +4,43 @@
...
+    multiply(value) {
+      result *= value;
+
+      return this;
+    },
+    divide(value) {
+      result /= value;
+
+      return this;
+    },
```
volodymyr-soltys97, line 37: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

nazarmatsevych (approved):
Good work!

PR #2235, opened 2024-04-29
Solution

raccoon1Andrew:


src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
...
+      this.result += num;
+
+      return this.result;
+    },
+    subtract(num) {
+      this.result -= num;
+
+      return this.result;
+    },
+    divide(num) {
```
lerastarynets, line 19: pls consider case with division to 0 and throw error in such case

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/makeCalculator.js
```diff
@@ -4,7 +4,45 @@
...
+      if (num === 0) {
+        return 'Division by zero';
```
anastasiiavorobiova, line 21:
```suggestion
        throw new Error('Division by zero');
```

anastasiiavorobiova (approved):
Great work!

PR #2238, opened 2024-04-30
add solution

Alenalenk:


src/makeCalculator.js
```diff
@@ -4,7 +4,34 @@
...
+    add(a) {
+      calculator.result += a;
+    },
+    subtract(a) {
+      calculator.result -= a;
+    },
+    multiply(a) {
+      calculator.result *= a;
+    },
+    divide(a) {
```
lerastarynets, line 18: consider case with division to 0, throw error in such case

lerastarynets (changes requested):

lerastarynets (changes requested):
/

anastasiiavorobiova (approved):
Awesome work!

PR #2240, opened 2024-04-30
Task make calculator

HigorPedroso:


src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
...
+      calculator.result *= number;
+
+      return this;
+    },
+
+    divide(number) {
+      calculator.result /= number;
+
+      return this;
+    },
```
EdPirro, line 30: Instead of having every callback edit the final `result` property you could simply return the result of the operation and update the `operate` callback to set result `this.result = operation(number)`.

This better follows the single responsibility principle by centralizing the `result` value update handling in `operate` and allowing each callback to simply perform the operation it's supposed to.

src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
...
+
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+
+    operate(operation, number) {
+      operation(number);
```
EdPirro, line 39: Look how you're always ignoring the callback return and assuming it will update `this.result`

EdPirro (approved):
Great job, @HigorPedroso! Keep it up! 🚀

I added a few comments about how you could change your solution to follow coding best practices, please have a look if you have the time

PR #2244, opened 2024-05-04
add solution

timaantonuk:


src/makeCalculator.js
```diff
@@ -3,8 +3,49 @@
...
+    },
+
+    multiply(a) {
+      this.result *= a;
+
+      return this;
+    },
+
+    divide(a) {
+      this.result /= a;
```
volodymyr-soltys97, line 30: It is necessary to add a check for 0, since it is not possible to divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

timaantonuk (commented):
fix

src/makeCalculator.js
```diff
@@ -3,8 +3,52 @@
...
+
+    multiply(a) {
+      this.result *= a;
+
+      return this;
+    },
+
+    divide(a) {
+      if (a === 0) {
+        throw new error('Can not divide by 0')
```
volodymyr-soltys97, line 31:
```suggestion
        throw new Error('Can not divide by 0')
```

volodymyr-soltys97 (changes requested):
Almost done!
Let's fix my comments to pass github tests 
image

src/makeCalculator.js
volodymyr-soltys97: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

VitaliyBondarenko1982 (approved):
Good job.

PR #2256, opened 2024-05-08
GG calculator

Teslaanazuma:
Create a makeCalculator function that returns an object that has the following fields:

Methods: add, subtract, multiply, divide, reset, operate.
The result property is initially 0.
How the calculator will work:

Each operate call takes a callback and a number and sets the appropriate value to the result property.
The reset method resets result value to 0.
add, subtract, multiply, divide are passed as callbacks to operate method.
The operate and reset methods can be called in a chain.

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -4,7 +4,50 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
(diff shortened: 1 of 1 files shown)


lerastarynets (approved):
You're making excellent progress!

PR #2260, opened 2024-05-09
Solution

laishchuk:


src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+      return b - a;
+    },
+
+    multiply(b, a) {
+      return b * a;
+    },
+
+    divide(b, a) {
+      return b / a;
+    },
```
volodymyr-soltys97, line 30: You need to add checking by 0, because you cannot divide by 0

src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    operate(callback, a) {
+      this.result = callback(this.result, a);
```
volodymyr-soltys97, line 11:
```suggestion
    operate(callback, value) {
      callback.call(result, value);
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/makeCalculator.js
laishchuk: divide(b, a) {
      if (a !== 0) {
        return b / a;
      }
    },

Correct checking?

src/makeCalculator.js
laishchuk: After this change my code working wrong. Can i keep my option?

PR #2264, opened 2024-05-10
solution

SrTrace:


src/makeCalculator.js
```diff
@@ -4,7 +4,33 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(value) {
+      calculator.result += value;
```
l4st1m0za, line 10: You should use 'this.result' instead of 'calculator.result' to make sure that 'result' is always accessed from the current instance of the object.

src/makeCalculator.js
```diff
@@ -4,7 +4,33 @@
...
+      calculator.result -= value;
+    },
+    multiply(value) {
+      calculator.result *= value;
+    },
+    divide(value) {
+      calculator.result /= value;
+    },
+    reset() {
+      calculator.result = 0;
```
l4st1m0za, line 22: Do not allow division by zero

l4st1m0za (changes requested):
Good job, but some changes needed

SrTrace:
Ok, I fixed remarks. Now looks better)

anastasiiavorobiova (approved):
Great work!

PR #2283, opened 2024-05-30
git commit -m 'add task solution'

iFedyna:


src/makeCalculator.js
```diff
@@ -5,6 +5,35 @@
...
+    },
+    subtract(num) {
+      result -= num;
+    },
+    multiply(num) {
+      result *= num;
+    },
+    divide(num) {
+      result /= num;
+    },
```
volodymyr-soltys97, line 25: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Good job 👍

PR #2285, opened 2024-06-01
solution

yaroslav1177:


src/makeCalculator.js
```diff
@@ -5,6 +5,39 @@
  */
 function makeCalculator() {
   // write code here
+  return {
+    result: 0,
```
roman-mirzoian, line 9: It is better to implement a getter for the ``result`` property.
This will ensure that the result value is accessed as a property and not as a variable.

Example:
```suggestion
function makeCalculator() {
  let result = 0;
  
  return {
```

src/makeCalculator.js
```diff
@@ -5,6 +5,39 @@
...
+
+    add(num) {
+      this.result += num;
+    },
```
roman-mirzoian, line 13: Methods must return `this` to allow chaining.
This will allow the methods to be called one after the other.
Fix it everywhere.

Example:
```suggestion
    add(num) {
      this.result += num;
      return this;
    },
```

src/makeCalculator.js
```diff
@@ -5,6 +5,39 @@
...
+      this.result -= num;
+    },
+
+    multiply(num) {
+      this.result *= num;
+    },
+
+    divide(num) {
+      this.result /= num;
+    },
```
roman-mirzoian, line 25: It is advisable to add a check for division by zero.
Example: 
```i
if (num === 0) {
  throw new Error('Cannot divide by zero!');
}
```

roman-mirzoian (changes requested):
Good job, but please pay attention to the comments.

PR #2287, opened 2024-06-03
Solution task

Karanelus:
How the calculator will work:
- Each `operate` call takes a callback and a number and sets the appropriate value to the `result` property.
- The `reset` method resets `result` value to 0.
- `add`, `subtract`, `multiply`, `divide` are passed as callbacks to `operate` method.
- The `operate` and `reset` methods can be called in a chain.

My solution yo this task

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -124,9 +124,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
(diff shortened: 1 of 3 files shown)


natalia-klonowska (changes requested):
Great work 🚀 just rename single word variables (a) for better readability

natalia-klonowska (approved):

PR #2288, opened 2024-06-06
Solution

AlinaMyronenko:


src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+
+    subtract(number) {
+      this.result -= number;
+    },
+
+    multiply(number) {
+      this.result *= number;
+    },
+
+    divide(number) {
```
lerastarynets, line 34: consider case with division to 0 and throw error in such case

lerastarynets (changes requested):

PR #2293, opened 2024-06-11
add task solution

ViktorZarvanskyi:


src/makeCalculator.js
```diff
@@ -4,7 +4,43 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(operand) {
+      this.result += operand;
+
+      return this;
```
anastasiia-tilikina, line 12: you don't need to return this from add subtract multiply and divide methods. cinse only operate and reset methods could be chained

anastasiia-tilikina (changes requested):
Well done!
just one mini comment

src/makeCalculator.js
ViktorZarvanskyi: done

VitaliyBondarenko1982 (approved):
GJ!

PR #2299, opened 2024-06-20
advanced_calculator

Anastasiia-Tatarchenko:


src/makeCalculator.js
```diff
@@ -3,8 +3,44 @@
 /**
  * @return {object}
  */
-function makeCalculator() {
-  // write code here
+function makeCalculator(operate) {
```
TarasHoliuk, line 6: This comment makes sense, please, fix it:

image

src/makeCalculator.js
```diff
@@ -3,8 +3,44 @@
...
+  const obj = {
```
TarasHoliuk, line 7:
```suggestion
  return {
```

no need to create variable here - you can immediately return it
BTW, obj is bad name 🙂

TarasHoliuk (changes requested):

TarasHoliuk (approved):

PR #2305, opened 2024-07-02
add solution

OlyaVidzi:


src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    multiply(operand) {
+      this.result *= operand;
+
+      return this;
+    },
+    divide(operand) {
+      this.result /= operand;
+
+      return this;
+    },
```
volodymyr-soltys97, line 28: You need to provide a check by 0, since you cannot divide by 0 and you need to throw an error
```suggestion
    divide(operand) {
      if (operand === 0) {
        throw new Error('Message error')
      } else {
        this.result /= operand;
      }

      return this;
    },
```

src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+    add(operand) {
+      this.result += operand;
+
+      return this;
+    },
+    subtract(operand) {
```
volodymyr-soltys97, line 14: Consider will be better if your fix it everywhere
```suggestion
    subtract(value) {
```

volodymyr-soltys97 (changes requested):
GJ!
Let's make your code better

lerastarynets (approved):
Keep up the great work!

PR #2311, opened 2024-07-08
Solution

iyehorova:


src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+      this.result -= value;
+    },
+
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      this.result = value === 0 ? 0 : this.result / value;
+    },
```
Teollan, line 36: Mathematically speaking, [division by zero is undefined](https://en.wikipedia.org/wiki/Division_by_zero), i.e. definitely not `0`. `NaN` is a better option. 

Alternatively, since js does not throw an error on division by zero, you can just not handle DbZ at all.

Teollan (approved):
Good job!

I`ll approve, but see one minor comment below.

src/makeCalculator.js
iyehorova: Agree, but I want to avoid any errors in this case. Ideally, no matter what number the user types, the application continues to work

ArturErmolenko (approved):
Nice 👍

PR #2314, opened 2024-07-10
add solution

RimmaBuzuluk:


src/makeCalculator.js
```diff
@@ -4,7 +4,46 @@
...
+    subtract(number) {
+      this.result -= number;
+
+      return this;
+    },
+    multiply(number) {
+      this.result *= number;
+
+      return this;
+    },
```
ArturErmolenko, line 24:
```suggestion
    result: 0,

    add(number) {
      this.result += number;

      return this;
    },
    
    subtract(number) {
      this.result -= number;

      return this;
    },
    
    multiply(number) {
      this.result *= number;

      return this;
    },
```

ArturErmolenko (approved):
GJ 👍 
Pay attention to correct formatting

PR #2315, opened 2024-07-11
Solution

YuliiaKosenchuk:


src/makeCalculator.js
```diff
@@ -4,7 +4,45 @@
...
+
+    multiply: function (valueResult) {
+      return function (value) {
+        return valueResult * value;
+      };
+    },
+
+    divide: function (valueResult) {
+      return function (value) {
+        return valueResult / value;
```
volodymyr-soltys97, line 30: It is necessary to add a check for 0, because 0 can be divided by according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/makeCalculator.js
```diff
@@ -4,7 +4,45 @@
...
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add: function (valueResult) {
+      return function (value) {
+        return valueResult + value;
+      };
+    },
```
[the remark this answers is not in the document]
YuliiaKosenchuk: I decided this - add: (currentResult) => (value) => currentResult + value, =)

src/makeCalculator.js
YuliiaKosenchuk: It wasn't in the tests, but it's helpful to do. thank you

src/makeCalculator.js
```diff
@@ -4,7 +4,31 @@
...
+
+    subtract: (currentResult) => (value) => currentResult - value,
+
+    multiply: (currentResult) => (value) => currentResult * value,
+
+    divide: (valueResult) => (value) => {
+      return value === 0 ? 'Error' : valueResult / value;
+    },
+
+    operate: function (callback, value) {
```
volodymyr-soltys97, line 20:
```suggestion
    operate(callback, value) {
```

src/makeCalculator.js
```diff
@@ -4,7 +4,31 @@
...
+      this.result = callback(this.result)(value);
+
+      return this;
+    },
+
+    reset: function () {
```
volodymyr-soltys97, line 26:
```suggestion
    reset() {
```

volodymyr-soltys97 (changes requested):
Almost done!

src/makeCalculator.js
YuliiaKosenchuk: when I write  => function in this place, the tests fail, I think it's related to using "This." Or just tests work that way.

operate: (callback, value) => {
      this.result = callback(this.result)(value);

      return this;
    },

src/makeCalculator.js
YuliiaKosenchuk: when I write  => function in this place, the tests fail, I think it's related to using "This." Or just tests work that way.

reset: () => {
      this.result = 0;

      return this;
    },

src/makeCalculator.js
```diff
@@ -4,7 +4,31 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add: (currentResult) => (value) => currentResult + value,
```
volodymyr-soltys97, line 10: You need to work with `this` fix it everywhere
```suggestion
    add(value) {
      this.result += value;
    }
```

src/makeCalculator.js
```diff
@@ -4,7 +4,31 @@
...
+
+    divide: (valueResult) => (value) => {
+      return value === 0 ? 'Error' : valueResult / value;
+    },
+
+    operate: function (callback, value) {
+      this.result = callback(this.result)(value);
+
+      return this;
+    },
```
volodymyr-soltys97, line 24:
```suggestion
    operate(callback, value) {
      callback.call(this, value);
      
      return this;
    },
```

volodymyr-soltys97: You read more about a call method here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

src/makeCalculator.js
```diff
@@ -4,7 +4,31 @@
...
+
+    reset: function () {
+      this.result = 0;
+
+      return this;
```
volodymyr-soltys97, line 29:
```suggestion
    reset() {
      this.result = 0;

      return this;
```

volodymyr-soltys97 (changes requested):
Let's fix your code by my comments

src/makeCalculator.js
YuliiaKosenchuk: this method I did not meet in training videos, thanks for the advice

YuliiaKosenchuk: this method I did not meet in training videos, thanks for the advice

src/makeCalculator.js
```diff
@@ -4,7 +4,40 @@
...
+      this.result -= value;
+    },
+
+    multiply(value) {
+      this.result *= value;
+    },
+
+    divide(value) {
+      if (value === 0) {
+        this.result = undefined;
```
volodymyr-soltys97, line 24:
```suggestion
        throw new Error('You cannot divide by 0')
```

volodymyr-soltys97 (approved):
Great 🔥

PR #2316, opened 2024-07-11
Solution

Kirill1908:
Solution

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -4,7 +4,49 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add: function (valueResult) {
+      return function (value) {
+        return valueResult + value;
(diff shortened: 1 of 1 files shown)


src/makeCalculator.js
```diff
@@ -4,7 +4,49 @@
...
+      return function (value) {
+        return valueResult + value;
+      };
+    },
+
+    subtract: function (valueResult) {
+      return function (value) {
+        return valueResult - value;
+      };
+    },
```
volodymyr-soltys97, line 20: Fix it everywhere
```suggestion
    subtract(value) {
      this.result /= value;
      
      return this;
    },
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix code by AI comments

lerastarynets (approved):
lgtm 🚀

PR #2317, opened 2024-07-11
solution

didarie:


src/makeCalculator.js
```diff
@@ -4,7 +4,33 @@
...
+    add(a) {
+      this.result += a;
+    },
+    subtract(a) {
+      this.result -= a;
+    },
+    multiply(a) {
+      this.result *= a;
+    },
+    divide(a) {
```
lerastarynets, line 18: consider case with devision to 0 and throw error in such case

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

anastasiiavorobiova (approved):
Great work!

PR #2322, opened 2024-07-13
Added solution

k-marchuk:


src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
...
+      return this;
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(value, result) {
```
ArturErmolenko, line 23: Let's imagine that this is a calculator and you want to add some number, you press +2, for example, the question is where will you take the result each time to pass it as the second arg?
Methods should take only one argument and take the result that is already stored in memory
In case of add you should increase result on value like `this.result += value`

src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
...
+      return result + value;
+    },
+
+    subtract(value, result) {
+      return result - value;
+    },
+
+    divide(value, result) {
```
ArturErmolenko, line 31: what happens if i pass 0 as value?)

ArturErmolenko (changes requested):
Good start, let's improve 🐝

src/makeCalculator.js
k-marchuk: There was no any mentions that these methods (`add, subtract, divide, multiply`) can be executed directly without `operate` method. Example code shows only `operate` case and all the tests passed, so I didn't expect any other cases.

src/makeCalculator.js
k-marchuk: `this.result `will be Infinity.

There was no any requirements about 'dividing by zero' case handling. Should I reset result to 0 or throw an error or something else?

BudnikOleksii: > `this.result `will be Infinity.
> 
> There was no any requirements about 'dividing by zero' case handling. Should I reset result to 0 or throw an error or something else?

Ok, it's just for a better user experience. Yes, you can throw an error or just notify a user that they can't divide by 0

BudnikOleksii (approved):
👍

PR #2328, opened 2024-07-17
add solution

Anya-Laban:


src/makeCalculator.js
```diff
@@ -4,7 +4,31 @@
...
+    add(num) {
+      this.result += num;
+    },
+    subtract(num) {
+      this.result -= num;
+    },
+    multiply(num) {
+      this.result *= num;
+    },
+    divide(num) {
```
lerastarynets, line 28: Consider case with division to 0 and throw error in such case

lerastarynets (changes requested):

etojeDenys (approved):
lgtm

PR #2329, opened 2024-07-17
v 1.0

maxproway:


src/makeCalculator.js
```diff
@@ -4,7 +4,51 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, value) {
+      this.result = callback(this.result, value);
```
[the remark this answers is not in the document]
anastasiia-tilikina: fix this comment

src/makeCalculator.js
```diff
@@ -4,7 +4,51 @@
...
+// console.log(calculator.result); // 0
+
+// calculator
+//   .operate(calculator.add, 10)
+//   .reset()
+//   .operate(calculator.subtract, 20)
+//   .operate(calculator.divide, 5)
+//   .operate(calculator.multiply, 7);
+
+// console.log(calculator.result); // -28
```
anastasiia-tilikina, line 54: remove this code

anastasiia-tilikina (changes requested):
Consider fixing ai comments, they are all relevant

maxproway:
this.result its what I did initially, but in this case I'm getting the error: 

```
...src\makeCalculator.js:10
      this.result += value;
      ^
```

let me PR my initial code.

maxproway:
Actually system doesn't allow me to do it, since  when I'm using this.result it didn't pass the test. Here is how my code looks like: 
image

src/makeCalculator.js
etojeDenys, line 54: remova all your comments

etojeDenys (approved):

src/makeCalculator.js
maxproway: Why did you approve this PR? Its not valid. I need a help with it please.

PR #2334, opened 2024-07-20
Solution

DariaNastas:


src/makeCalculator.js
```diff
@@ -4,7 +4,44 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add: function (value) {
```
etojeDenys, line 9:
```suggestion
    add(value) {
```
we can simplify that

etojeDenys (approved):

src/makeCalculator.js
DariaNastas: Thanks! now that's how I'll write ;)

PR #2338, opened 2024-07-22
solution

adonch:


src/makeCalculator.js
```diff
@@ -4,7 +4,33 @@
...
+      return this;
+    },
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(currentResult, number) {
+      return currentResult + number;
+    },
```
volodymyr-soltys97, line 21: Fix it everywhere
```suggestion
    add(number) {
      this.result += number;
      
      return this;
    },
```

src/makeCalculator.js
```diff
@@ -4,7 +4,33 @@
...
+    subtract(currentResult, number) {
+      return currentResult - number;
+    },
+    multiply(currentResult, number) {
+      return currentResult * number;
+    },
+    divide(currentResult, number) {
+      return currentResult / number;
+    },
```
volodymyr-soltys97, line 30: It is necessary to add a check for 0, since it is not possible to divide by 0 according to the rules

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/makeCalculator.js
adonch: Але якщо я так зроблю, то в мене перестане все працювати

etojeDenys (approved):
lgtm

PR #2346, opened 2024-07-29
Solution

EduarfF:


src/makeCalculator.js
```diff
@@ -4,7 +4,49 @@
...
+  function subtract(value) {
+    result -= value;
+  }
+
+  function multiply(value) {
+    result *= value;
+  }
+
+  function divide(value) {
+    result /= value;
```
SarakhmanFTMentor, line 22: what about zero division?

SarakhmanFTMentor (approved):
you can improve your solution if you want :)

PR #2347, opened 2024-08-01
Solution

Marta-Matsevych:


src/makeCalculator.js
```diff
@@ -4,7 +4,28 @@
...
+
+  return {
+    get result() {
+      return result;
+    },
+
+    add: (number) => result + number,
+    subtract: (number) => result - number,
+    multiply: (number) => result * number,
+    divide: (number) => result / number,
```
lerastarynets, line 17: Consider case with division to 0 and throw error in such case

lerastarynets (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #2357, opened 2024-08-12
add task solution

ipaiareli:


src/makeCalculator.js
```diff
@@ -4,7 +4,30 @@
...
+      return a + b;
+    },
+    subtract(a, b) {
+      return a - b;
+    },
+    multiply(a, b) {
+      return a * b;
+    },
+    divide(a, b) {
+      return a / b;
```
volodymyr-soltys97, line 19: You need to add a check for division by 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's fix your code by AI comments

volodymyr-soltys97 (changes requested):
Hi, you need to remove duplicate methods

etojeDenys (approved):
good job

PR #2359, opened 2024-08-13
calculator solution

MaksymMohyla:


src/makeCalculator.js
```diff
@@ -4,7 +4,36 @@
...
+    },
+    subtract(num) {
+      calculator.result -= num;
+    },
+    multiply(num) {
+      calculator.result *= num;
+    },
+    divide(num) {
+      calculator.result /= num;
+    },
```
volodymyr-soltys97, line 21: It is necessary to add a check by 0, because according to the rules, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    subtract(num) {
+      calculator.result -= num;
+    },
+    multiply(num) {
+      calculator.result *= num;
+    },
+    divide(num) {
+      if (num === 0) {
+        console.error('Division by zero is not allowed');
```
volodymyr-soltys97, line 21: Don't need to use console, because now tests are failed
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw
```suggestion
        throw new Error('Division by zero is not allowed');
```

lerastarynets (approved):
lgtm 🚀

PR #2372, opened 2024-08-23
Added logic

CamilleHaus:


pedro-ruas (approved):
Camille, I believe you made the key steps very well.
I advise to take a look on the AI Mentor analysis, as it is flawless for this particular code and those points it says can be improved in the future.

PR #2373, opened 2024-08-23
Solution

KseniiaKrytska:


Oleksii-Siviolov (approved):
Well done. The function would operate, but I would suggest some improvements:
1. I would keep all methods return 'this' to support method chaining.
2. Refactor the 'operate' like this:
```
operate(callback, number) {
      return this;
    },
 ```
 to pass the current result as an argument to the callback for more flexibility.

PR #2374, opened 2024-08-24
solved

Artem-Brui:
Solved

### DIFF
--- a/src/makeCalculator.js
+++ b/src/makeCalculator.js
@@ -5,6 +5,45 @@
  */
 function makeCalculator() {
   // write code here
+  return {
+    result: 0,
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    add(number) {
+      this.result += number;
+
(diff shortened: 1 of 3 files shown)


src/makeCalculator.js
```diff
@@ -5,6 +5,45 @@
...
+    },
+
+    multiply(number) {
+      this.result *= number;
+
+      return this;
+    },
+
+    divide(number) {
+      this.result /= number;
```
volodymyr-soltys97, line 36: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/makeCalculator.js
Artem-Brui: Thanks, solved.

VitaliyBondarenko1982 (approved):
Looks good

PR #2380, opened 2024-08-27
Solution  make calculator

OlenaZiniak:


src/makeCalculator.js
```diff
@@ -4,7 +4,33 @@
...
+      calculator.result += x;
+    },
+    subtract: function (x) {
+      calculator.result -= x;
+    },
+    multiply: function (x) {
+      calculator.result *= x;
+    },
+    divide: function (x) {
+      calculator.result /= x;
```
volodymyr-soltys97, line 19: You need to add a check for 0, because you cannot divide by 0

volodymyr-soltys97 (changes requested):
GJ!
Let's fix your code by AI comments

volodymyr-soltys97 (approved):
GJ!

PR #2386, opened 2024-08-29
Solution

hma-3:


src/makeCalculator.js
```diff
@@ -4,7 +4,57 @@
...
+        case this.multiply: {
+          this.multiply(number);
+          break;
+        }
+
+        case this.divide: {
+          this.divide(number);
+          break;
+        }
+      }
```
Moroz-Dmytro, line 53: Try to make it shorter using `call` - https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Function/call

Moroz-Dmytro (changes requested):
Good work 👍 
But let's make your solution better

BudnikOleksii (approved):
Good work 👍

PR #2387, opened 2024-08-31
Solution

19Eduard99:


src/makeCalculator.js
```diff
@@ -4,7 +4,49 @@
...
+
+  const subtract = (prevResult, num) => {
+    return prevResult - num;
+  };
+
+  const multiply = (prevResult, num) => {
+    return prevResult * num;
+  };
+
+  const divide = (prevResult, num) => {
```
volodymyr-soltys97, line 27: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code
You need to fix your code by AI comments
Also, you can immediately return the object with the necessary methods

etojeDenys (approved):
lgtm

PR #2395, opened 2024-09-03
solution

YuliiaShatkovska:


src/makeCalculator.js
```diff
@@ -4,7 +4,44 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
```
KrivorotkoYurii, line 7: Nicely done!)🔥 You can also do this task by creating just one object and including result as a key of the object. I like your version, it is even shorter)

KrivorotkoYurii (approved):
Good Job!🔥🔥🔥

Moroz-Dmytro (approved):
👍

PR #2396, opened 2024-09-03
Solution

KrivorotkoYurii:


src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
 'use strict';
 
-/**
- * @return {object}
- */
 function makeCalculator() {
-  // write code here
+  const resultingObject = {
```
Valeraom: 1.  [NAMING] - Use proper names for variables. If function was called `makeFruit`, it should return `fruit`

src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
...
+    result: 0,
+    add(a) {
+      resultingObject.result += a;
+
+      return resultingObject.result;
```
Valeraom: Your solution works without returnings

src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
...
+    },
+    reset() {
+      resultingObject.result = 0;
+
+      return this;
+    },
+    operate(callback, x) {
+      resultingObject.result = callback(x);
```
Valeraom: So, you can just call this function without assignment

Valeraom (changes requested):
Good job!
There are some corrections.

And I'm not sure about this solution. I would change `result` using `this` and `.bind()` in `operate` to connect context. It's just my opinion, maybe your variant is good. The mentor will check it)

src/makeCalculator.js
```diff
@@ -1,10 +1,33 @@
...
+      calculator.result -= a;
+    },
+    multiply(a) {
+      calculator.result *= a;
+    },
+    divide(a) {
+      calculator.result /= a;
+    },
+    reset() {
+      calculator.result = 0;
```
Moroz-Dmytro, line 19:
```suggestion
      this.result += a;
    },
    subtract(a) {
      this.result -= a;
    },
    multiply(a) {
      this.result *= a;
    },
    divide(a) {
      this.result /= a;
    },
    reset() {
      this.result = 0;
```

Moroz-Dmytro (approved):

PR #2400, opened 2024-09-03
Solution

Valeraom:


src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
...
+    },
+
+    divide(num) {
+      this.result /= num;
+    },
+
+    operate(operation, num) {
+      const operationContext = operation.bind(this);
+
+      operationContext(num);
```
BudnikOleksii, line 29: You can use `call` instead as you use this function immediately and only once

BudnikOleksii (approved):
Good job 👍

PR #2404, opened 2024-09-06
solution

Ostkreuzzz:


Svidruk (approved):
Good job ! 
But mentors may complain about using an arrow function as a method.

src/makeCalculator.js
```diff
@@ -4,7 +4,26 @@
...
-  // write code here
+  return {
+    result: 0,
+
+    add: (a, b) => a + b,
+    subtract: (a, b) => a - b,
+    multiply: (a, b) => a * b,
+    divide: (a, b) => a / b,
+
+    operate(callback, number) {
```
polosanya, line 15: Better use one way of creating methods

polosanya (approved):
👍

PR #2407, opened 2024-09-08
add solution

Tulenty:


src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
...
+      result *= value;
+      return this;
+    },
+
+    divide(value){
+      if(value !== 0){
+        result /= value;
+        return this;
+      } else {
+        console.log('Error: Division by zero');
```
volodymyr-soltys97, line 30:
```suggestion
        throw new Error('Error: Division by zero');
```

src/makeCalculator.js
```diff
@@ -4,7 +4,47 @@
...
  */
 function makeCalculator() {
-  // write code here
+  let result = 0;
+
+  return {
+    add(value){
+      result += value;
+      return this;
+    },
```
volodymyr-soltys97, line 13: Don't forget about code styles
```suggestion
    add(value) {
      result += value;
      
      return this;
    },
```

volodymyr-soltys97 (approved):
Good job 👍

PR #2408, opened 2024-09-08
Develop

gnat-vasylevych:


src/makeCalculator.js
```diff
@@ -5,6 +5,39 @@
...
+
+    subtract(num) {
+      this.result -= num;
+    },
+
+    multiply(num) {
+      this.result *= num;
+    },
+
+    divide(num) {
```
lerastarynets, line 23: Consider case with division to 0 and throw error in such case

lerastarynets (changes requested):

src/makeCalculator.js
```diff
@@ -5,6 +5,43 @@
  */
 function makeCalculator() {
   // write code here
```
volodymyr-soltys97, line 7:
```suggestion
```

volodymyr-soltys97 (approved):
Good job! 👍

PR #2420, opened 2024-09-25
task was solved

NazariiAlieksieiev:


src/makeCalculator.js
```diff
@@ -2,9 +2,34 @@
...
+  return {
+    result: 0,
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+    add(y, x) {
+      return y + x;
+    },
```
volodymyr-soltys97: Fix it everywhere
```suggestion
    add(value) {
      this.result += value;
    },
```

src/makeCalculator.js
```diff
@@ -2,9 +2,34 @@
...
+      return y - x;
+    },
+    multiply(y, x) {
+      return y * x;
+    },
+    divide(y, x) {
+      return y / x;
+    },
+    operate(callback, param) {
+      this.result = callback(this.result, param);
```
volodymyr-soltys97: You can rewrite like this
```suggestion
      callback.call(this, param);
```

src/makeCalculator.js
```diff
@@ -2,9 +2,34 @@
...
+    },
+    subtract(y, x) {
+      return y - x;
+    },
+    multiply(y, x) {
+      return y * x;
+    },
+    divide(y, x) {
+      return y / x;
+    },
```
volodymyr-soltys97: You need to add a check for 0, because according to the rules of mathematics, you cannot divide by 0

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Good job! 👍

PR #2470, opened 2024-10-31
Solution

MakksymSly:


src/makeCalculator.js
```diff
@@ -4,7 +4,30 @@
...
+    result: 0,
+    add: (num) => (calculator.result += num),
+    subtract: (num) => (calculator.result -= num),
+    multiply: (num) => (calculator.result *= num),
+    divide: (num) => {
+      if (num === 0) {
+        return;
+      }
+      calculator.result /= num;
+    },
```
ab3MN, line 17: It would be a good practice to use `this `instead of object's name

ab3MN (approved):
Good job👍
But try to change object's name(calculator) to `this`

src/makeCalculator.js
BudnikOleksii: > It would be a good practice to use `this `instead of object's name

Good point

BudnikOleksii (approved):

PR #2481, opened 2024-11-03
Solution

chelsea7smile:


ab3MN (approved):
Good job 👍
It would be better to save all the methods in object and use `this` instead calculator in function and call operation in operate with `call` like :

```
 operation.call(this, num);
 return this;
 ```

chelsea7smile:
> Good job 👍 It would be better to save all the methods in object and use `this` instead calculator in function and call operation in operate with `call` like :
> 
> ```
Yeah, it looks better and improve the code style, txh mate, I fix it!

BudnikOleksii (approved):
👍

PR #2482, opened 2024-11-03
Solution

iirk1:


src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
...
+      return this;
+    },
+
+    multiply(value) {
+      this.result *= value;
+      return this;
+    },
+
+    divide(value) {
+      this.result /= value;
```
volodymyr-soltys97: You need to add a check for 0, because you cannot divide by 0 according to the rules of mathematics

src/makeCalculator.js
```diff
@@ -1,10 +1,41 @@
...
- * @return {object}
- */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(value) {
+      this.result += value;
+      return this;
```
volodymyr-soltys97: Add the blank line before the return, fix it everywhere
```suggestion
      this.result += value;
      
      return this;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
GJ!

PR #2490, opened 2024-11-04
solution

maxsabo:


src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  return {
+    result: 0,
+
+    add(a, b) {
```
schyrva: you can modify these methods to work with `this.result` and a single parameter, as `operate` should handle updating `result`. Like this `add(number) {this.result += number;}`.

src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
...
+
+    subtract(a, b) {
+      return a - b;
+    },
+
+    multiply(a, b) {
+      return a * b;
+    },
+
+    divide(a, b) {
```
schyrva: divide method currently returns `a` if `b` is `0`, which may lead to confusion. A better approach might be to set result to `Infinity`, `null`, or log an `error message`, clearly indicating that division by zero is not allowed.

schyrva (approved):
Good Job! 👍 
But, you can make few adjustments to make it better :)

schyrva:
Great Job! 👍 👍

BudnikOleksii (approved):

PR #2550, opened 2025-01-06
Calculator solution

aCTORr8:


src/makeCalculator.js
```diff
@@ -4,7 +4,43 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+
+    add(x) {
+      calculator.result += x;
```
etojeDenys:
```suggestion
      this.result += x;
```
it's much more reusable to use `this` in such cases

etojeDenys (approved):
good job

PR #2601, opened 2025-02-17
add task solution

AnOvramenko:


daniil-trusov (approved):
Awesome job! Your solution is compact, simple and easy to read, keep it up!

PR #2606, opened 2025-02-17
add task solution

maponomarenko:


radykhovska (approved):
Exellent! everything is ok)

PR #2626, opened 2025-03-06
create calculator

VladStep186:


src/makeCalculator.js
```diff
@@ -4,7 +4,45 @@
...
+    },
+
+    multiply(num) {
+      this.result *= num;
+
+      return this;
+    },
+
+    divide(num) {
+      this.result /= num;
```
maxim2310: need to add check if number is not zero

maxim2310 (changes requested):

etojeDenys (approved):

PR #2658, opened 2025-03-28
feat: add task solution

s1rserg:


s1rserg:
@mateacademy-ai-mentor 
Would the implementation be better if I gave to the add subtract etc methods when I call them in operate `this` object and use `this.result` or now when I give result property as a parameter its still good?

PR #2694, opened 2025-04-24
add solution

lidiia-tsymborovych:


src/makeCalculator.js
```diff
@@ -4,7 +4,43 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const CALCULATOR = {
```
DenisPitsul, line 7: Just a recommendation:
You don't have to create variable of object here. You can return object:
`return {...}`

DenisPitsul (approved):
Good job!

PR #2695, opened 2025-04-25
add advanced calculator solution

DenisPitsul:


src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
...
+  return {
+    result: 0,
+
+    operate(operation, value) {
+      this.result = operation(this.result, value);
+
+      return this;
+    },
+
+    add(a, b) {
```
ArturErmolenko:
```suggestion
    add(value) {
      this.result += value
      return this;
    }
```
that way calculator will work correctly

ArturErmolenko (changes requested):
According to the task requirements (calculator.operate(calculator.add, 21)), your add, subtract, and other methods should be methods that accept only one argument (a number) and modify this.result internally.

In other words, add and subtract should work directly with this.result.

However, in your implementation, the operations have become regular functions that don't know anything about this — they just take two numbers as arguments.

ArturErmolenko (approved):
Nice 👍

PR #2879, opened 2025-10-23
solu

7uckingbad:


src/makeCalculator.js
```diff
@@ -5,6 +5,35 @@
...
+    add: function (a, b) {
+      return a + b;
+    },
+    subtract: function (a, b) {
+      return a - b;
+    },
+    multiply: function (a, b) {
+      return a * b;
+    },
+    divide: function (a, b) {
```
maxim2310, line 20: b possibly could be 0

maxim2310 (changes requested):

PR #2915, opened 2025-12-08
feat: ✨ implemented calculator functionality

de-vasta:


src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    multiply(arg) {
+      this.result *= arg;
+
+      return this;
+    },
+    divide(arg) {
+      this.result /= arg;
+
+      return this;
+    },
```
DemonWayne, line 29: Those returns are unnecessary cause we use those functions like callbacks and don't expect them to return anything, so we can remove those returns in my opinion.

src/makeCalculator.js
```diff
@@ -4,7 +4,39 @@
...
+    multiply(arg) {
+      this.result *= arg;
+
+      return this;
+    },
+    divide(arg) {
+      this.result /= arg;
```
DemonWayne: I think that raw divide could be harmful and break all calculations if arg is **zero**. We would probably get infinity, and I don't think that's an expected behavior for this calculator. Maybe you can check if arg is **zero** and reset the result?

DemonWayne (changes requested):
Looks good, except for the specified problems below 👀

src/makeCalculator.js
de-vasta: You got a point. We do not really need to when we don't call them directly, but within `operate` function

src/makeCalculator.js
de-vasta: Ok. In general, as far as I know, calculators tend to give you `NaN`, `Infinity` or some message like "Cannot divide by zero" or so after dividing by zero. You provided quite the interesting approach

de-vasta (commented):
Removed method chaining where it's not used in tests

src/makeCalculator.js
```diff
@@ -8,23 +8,20 @@ function makeCalculator() {
...
       this.result *= arg;
-
-      return this;
     },
     divide(arg) {
-      this.result /= arg;
+      if (!arg || arg === 0) {
+        this.reset();
+        throw new Error('Cannot divide by zero');
+      }
```
de-vasta, line 26: @DemonWayne, I provided zero handling in division

src/makeCalculator.js
```diff
@@ -4,7 +4,36 @@
...
+      this.result += arg;
+    },
+    subtract(arg) {
+      this.result -= arg;
+    },
+    multiply(arg) {
+      this.result *= arg;
+    },
+    divide(arg) {
+      if (!arg || arg === 0) {
```
DemonWayne: I think you need either only to check `!arg` or `arg === 0`, like `!arg` by default already checks if the value is `0`

DemonWayne (approved):
Looks good! Specified a minor thing in the if check, but don't think this really matter.

src/makeCalculator.js
de-vasta: well indeed. `0` is about the same as `false`

DemonWayne (approved):

ArturErmolenko (approved):

PR #2922, opened 2025-12-14
Advanced calculator

illia-001:
Solution


src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
...
+
+      return calculator.result;
+    },
+    reset() {
+      calculator.result = 0;
+
+      return this;
+    },
+    operate(callback, number) {
+      calculator.result = callback(number);
```
Dimononon: result of operation assinged to `result` 2 times. First time inside arithmetic operation and second time directly from callback return

src/makeCalculator.js
```diff
@@ -4,7 +4,41 @@
  * @return {object}
  */
 function makeCalculator() {
-  // write code here
+  const calculator = {
+    result: 0,
+    add(number) {
+      calculator.result += number;
```
Dimononon: fields from parent object should be called using keyword `this`
Same to all below

Dimononon (changes requested):
Almost, need some adjustments)

Dimononon (approved):
Good job

PR #2931, opened 2025-12-21
Solution

dmytro-halieba:


src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+
+    divide(number) {
+      return this.result /= number;
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
```
DemonWayne, line 34: Those returns are unnecessary 'cause we use those functions as callbacks and don't really expect them to return anything, so I think we can remove those returns.

src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+    subtract(number) {
+      return this.result -= number;
+    },
+
+    multiply(number) {
+      return this.result *= number;
+    },
+
+    divide(number) {
+      return this.result /= number;
```
DemonWayne: I think that raw divide could be harmful and break all calculations if number is zero. We would probably get infinity, and that's maybe not an expected behavior for this calculator. Could you check if the number is zero and maybe **throw an error** or `reset` the result?

src/makeCalculator.js
```diff
@@ -4,7 +4,37 @@
...
+    },
+
+    reset() {
+      this.result = 0;
+
+      return this;
+    },
+
+    operate(callback, number) {
+      this.result = callback.call(this, number);
```
DemonWayne: And here we don't need this assigning, we already assign value in the called callbacks

DemonWayne (changes requested):
Looks okay, specified some minor problems below

src/makeCalculator.js
DemonWayne: Oh, don't mind that I also selected reset here; ignore this comment about reset.

src/makeCalculator.js
dmytro-halieba: fixed in [90be0ae](https://github.com/mate-academy/js_advanced_calculator/pull/2931/commits/90be0ae9d1d0b3d8d7e81e5afb6d2d4e4895bed4)

src/makeCalculator.js
dmytro-halieba: fixed in [90be0ae](https://github.com/mate-academy/js_advanced_calculator/pull/2931/commits/90be0ae9d1d0b3d8d7e81e5afb6d2d4e4895bed4)

PR #2994, opened 2026-03-12
Solution

protocolo23:
contém uma implementação simples de um calculadora em JavaScript.


protocolo23:
I understood the method and the content, but I still have some questions:

Chaining:  
Why do I need to return this in each method? Could this cause problems in some cases?

Operate method:  
What is the advantage of using operate instead of calling add, subtract directly?

Design:  
Is it better to build the calculator with a class or with a function that returns an object? What’s the difference?

Validation:  
What happens if I pass invalid values (like null or a string)? Should I handle this?

Division by zero:  
How should I deal with divide(0)? Is it better to throw an error or just let it return Infinity?

Export:  
In module.exports, is it better to export the makeCalculator function or an instance of it?

Testing:  
How can I create tests to make sure each operation works correctly? Is it worth using Jest or Mocha?

raulriato (changes requested):
Your code loks fine, but you have a critical problem. You're disabling `eslint`. this is not ok. You should work with it to make clean code.

protocolo23:
> Seu código funciona bem, mas você tem um problema crítico. Você está incapacitando. Isso não está certo. Você deve trabalhar com ele para criar código limpo. eslint

Do I just need to remove the line I commented out from ESLint, or do I have to change the logic?

IanLuan (approved):
Great Job!

PR #3008, opened 2026-04-02
Solution

timurradkevic:


Tysiachnyk (approved):
Hello! You've done a great job! 
My solution is a bit different. 
I'm not sure now which one is better. 
It's worth thinking about.
Thanks!

PR #3012, opened 2026-04-05
Solution

Tysiachnyk:


timurradkevic (changes requested):
Problem: arithmetic methods take result as a parameter instead of using this.result, and operate does not preserve the correct this context. Methods should update this.result directly and support chaining.

Tysiachnyk:
> Problem: arithmetic methods take result as a parameter instead of using this.result, and operate does not preserve the correct this context. Methods should update this.result directly and support chaining.

Thanks for the review! I’ve fixed it, and thanks to that, I learned about the 'call' method.

In my previous solution, I tried to separate responsibilities: the arithmetic methods handle calculations, while 'operate' handles writing. I thought that the 'operate' method had access to the 'this' keyword and the chaining seemed working. 
However, I might not fully understand it yet — it’s something worth thinking about.

I’m resubmitting it for review.

Thanks so much!

